PR #15, opened 2018-09-14
Define xrange in python3

cclauss:
__xrange(i)__ was removed in Python 3 in favor of __range(i)__ so for small values of __i__, use __range()__ in both versions of Python and for larger values of __i__ define __xrange()__ in Python 3 to be identical to __range()__.

* Also rename __typeError__ --> [__TypeError__](https://docs.python.org/3/library/exceptions.html#TypeError)


cclauss:
The remaining undefined names in Python 3 are covered in #16.

hoffmansc:
The relevant files here are actually written by the original author and most of them are not necessary for our code. I recommend we do a couple things:
* delete the unnecessary files (everything except `__init__.py`, `commands.py`, `predict_lr.py`, `train_pr.py`, `fadm/lr/*`, and `fadm/util/*` from what I can tell)
* make the `TypeError` fix in `fadm/lr/pr.py`
* make note of the above changes in `prejudice_remover.py` in the section which discusses the fairness-comparison changes
* add the flake tests to Travis in addition to our pytest tests (it looks like these aren't being run here)
  * you may need to add the flake lines to both `before_script` sections and add back the `exclude` section (this seems to be a bug in Travis -- it spawns 4 builds otherwise)

cclauss:
I believed that the four runs were a featured not a bug.  Two runs do flake8 static analysis while in parallel two other runs do pytest.  By doing it all in parallel, we fail faster.

Also, perhaps it is better for you to delete the unnecessary files in a separate PR first.  Then after that PR is merged, we can retool this PR around those changes

hoffmansc:
Okay, I think I understand. Are you able to run a different script for different builds then? It looks like they all skip pytest right now. Also, is it necessary to install the pip requirements for the flake tests to run? If so, I'm not sure it saves any time to run in parallel. However, it might still be useful to run both flake and pytest separately so flake errors don't cause pytest not to run

BTW, I submitted a PR (#26) which removes the files

PR #21, opened 2018-09-18
Added algorithms and test file for Celis et al. paper.

vijaykeswani:
As per discussion with Sameep Mehta, I have added support for the Fair Classification meta-algorithm that we developed. The implementation is based on the following work: 
_Classification with Fairness Constraints: A Meta-Algorithm with Provable Guarantees_ - https://arxiv.org/abs/1806.06055.

The algorithm we suggest in the paper takes the fairness metric as input and corresponding to the metric specified, the optimization problem is different. In this implementation, we provide code for optimizing two fairness metrics: Statistical Rate and False Discovery Rate. 

The main implementation file is **meta_fair_classifier.py**. The algorithm is implemented in the folder **celisMeta**. I have added a Jupyter test notebook for this algorithm as well - **demo_meta_classifier.ipynb**. 

I have tried to loosely follow the coding style used for other implemented algorithms. Any comments on the code/algorithm will be very helpful and please let me know any changes I should make before this can be merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Get in touch with us on [Slack](https://aif360.slack.com) (invitation [here](htt
 * Calibrated Equalized Odds Postprocessing ([Pleiss et al., 2017](https://papers.nips.cc/paper/7151-on-fairness-and-calibration))
 * Learning Fair Representations ([Zemel et al., 2013](http://proceedings.mlr.press/v28/zemel13.html))
 * Adversarial Debiasing ([Zhang et al., 2018](http://www.aies-conference.com/wp-content/papers/main/AIES_2018_paper_162.pdf))
+* Meta-Algorithm for Fair Classification ([Celis et al.. 2018](https://arxiv.org/abs/1806.06055))
 
 ## Supported fairness metrics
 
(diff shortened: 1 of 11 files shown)


hoffmansc:
It looks like the demo notebook is taking a too long to run. We could up the time limit but I don't think we should have such a lengthy unit test anyway. I propose we do a few things:
* keep the notebook but remove the test_meta_classifier.py file
* add a short unit test similar to test_disparate_impact_remover.py which tests basic usage and corner cases and such (should only run `fit` and `predict` one or two times since these are expensive)

Also, a few other things:
* there's a typo -- `MetaFairClassifer` should be `MetaFairClassifier`
* if you could add a line describing the new arg `type` in `__init__` i.e. state the valid values ('fdr', 'sr')
* it seems that there are some unused variables in meta_fair_classifier.py like `class_attr`, `y_test`, and `x_control_test` and some variables like `sensitive_attr` are assigned default values after they've been used
* the relative imports for files in the `celisMeta/` folder need to look like `from .General import *` and `from . import utils as ut` to work with Python 3
* you can add a link to your paper in the main README under "Supported bias mitigation algorithms"

Thanks for staying patient

vijaykeswani:
Thanks a lot for the review. I made the suggested changes and created a new test file instead of the earlier one based on the notebook. Its passed the tests now.

Please let me know if there's any other change required before it can be merged.

PR #27, opened 2018-09-21
Generate MEPS data using R

ckadner:
In order to run the [Medical Expenditure Tutorial](https://github.com/IBM/AIF360/blob/master/examples/tutorial_medical_expenditure.ipynb) one needs to download the [MEPS](https://github.com/IBM/AIF360/blob/master/aif360/data/raw/meps/README.md) datasets and convert them into CSV files using SAS or SPSS. Since neither SAS or SPSS are free, I propose to use a R script to do the download and conversion.

### DIFF
--- a/aif360/data/raw/meps/README.md
+++ b/aif360/data/raw/meps/README.md
@@ -14,20 +14,58 @@ The Medical Expenditure Panel Survey (MEPS) data consists of large scale surveys
 
 ## Download instructions
 
-In order to use the MEPS datasets with AIF360, please follow the following directions to download the datafiles and convert into csv files. 
+In order to use the MEPS datasets with AIF360, please follow the following directions to download the datafiles and 
+convert into csv files. 
 
-While the instructions below require the use of SPSS, instructions/code on using alternatives, including R, SAS, and Stata, are available at the [AHRQ MEPS Github repository](https://github.com/HHS-AHRQ/MEPS).
+Follow either set of instructions below for using R or SPSS. Further instructions for SAS, and Stata, are available at 
+the [AHRQ MEPS Github repository](https://github.com/HHS-AHRQ/MEPS).
 
-1. 2015 full Year Consolidated Data File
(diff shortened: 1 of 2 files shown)


ckadner:
@monindersingh -- since you worked on the MEPS notebook and download instruction, would you mind doing the review?

I did compare the resulting CSV files with the ones produced by SPSS. The files are identical except for the use of scientific notation for some large integer values (`100000` vs `1e+05`) which should not matter since the data will be loaded into a Pandas DataFrame.

ckadner:
Just FYI, I did try to do the conversion using various Python libraries:
- `savReaderWriter`
- `pyreadstat`
- `sas7bdat_converter`

but neither of them produced equivalent CSV files, if it worked at all.

------
Using Pandas `read_sas()` method worked for loading the `.ssp` files but it:
- produced a vastly larger data frame
- all `int64` columns came out `float64` *
- some columns had non-0 values where the SPSS generated CSV files had zeros **

(*) The `int64` vs `float64` should not be a functional issue but looks odd in the notebook output. Converting more than 1900 column from `float64` to `int64` took more than 10 mins but since only 138 features are retained this could be considered in the future.

(**) there may be some floating point errors where Pandas generate `5.39760534693e-79` instead of `0.0`

monindersingh:
Also, @ckadner, can you please add the following line "The user is responsible for reading and abiding by any copyright/usage rules and restrictions on the MEPS web site before running the script." right after the line "**Generating CSV files with R**" in the README file. 
Also, please include a print command with this message in the generate_data.R file.

ckadner:
@monindersingh -- I added a "Data Use Agreement" section to the README and a notice at the top of the script which has to be confirmed by the user running the script.

@nrkarthikeyan -- I added a brief comment at the beginning of the R script describing its function and I added the CRAN download link to the README. However I left the Homebrew notice for Mac users since it is the simplest approach to get this script running. R users will already have R installed and non-R users should not have to do a lot of work just to get the MEPS data (IMO).

PR #28, opened 2018-09-24
Update Notebook for dataset location and package install.

scottdangelo:
The location of the German credit data has moved. Update the
means of downloading it.
Use the pypi package for aif360 to download it.
Add logic to download the data to proper directory.
Add install of `cvxpy` to notebook.
Change the epsilon for the mitigation to the proper value of 0.1
Remove output and execution count.
Add Random seeding for consistent test output.

### DIFF
--- a/examples/tutorial_credit_scoring.ipynb
+++ b/examples/tutorial_credit_scoring.ipynb
@@ -24,10 +24,12 @@
     "\n",
     "\n",
     "\n",
-    "![image](images/Complex_NoProc_V3.jpg)\n",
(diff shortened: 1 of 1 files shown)


hoffmansc:
There seems to be a discrepancy with the mean difference value after debiasing. I'm getting -0.136480 vs 0.018250 as shown in the static version. We're still investigating what changed to cause this.

hoffmansc:
No longer relevant since we use Reweighing now instead of OptimPreproc.

Issue #35, opened 2018-10-11
Python 3.6 failing


michaelhind:
The latest PR failed for python 3.6 (but passed for 2.7).  Here's the details: https://travis-ci.com/IBM/AIF360/jobs/151221128

nrkarthikeyan:
Seems like it passed after i restarted the build.

hoffmansc:
#37 fixes this

hoffmansc:
#37 did not fix this...

michaelhind:
Looks like merge of issue #40 fixed this.

PR #40, opened 2018-10-23
fixed truncated division error

hoffmansc:
was causing problems in gradient descent
also made StandardDataset features order consistent


hoffmansc:
Fix for #35. Seems to make it consistent across Python versions now

PR #55, opened 2018-11-15
README.md breaking very line lines to ~80 cols

adrinjalali:
Reading the README in a text editor without line wrapping and/or on a wide monitor seems odd with the current long lines.

This PR only wraps the lines to ~80 chars. The rendered output is no different than the current in master.


michaelhind:
Thanks for the contribution.  This will also help with diffs.

PR #59, opened 2018-12-04
opt_tools.py fixed to use cvxpy new imports

mariaborbones:
I have updated the opt_tools.py file in order to use the updated cvxpy functions.


adrinjalali:
I guess it also makes sense to include some tests against the latest release of that package to detect the issues in the CI as they happen.

Issue #43, opened 2018-11-01
features_to_drop in Class  MEPSDataset19 not working


sainyam:
I tried to run the MEPSDataset19 example with features_to_drop. It ran with some features but not with some eg. 'PHQ242'

Error it gave: KeyError: "['PHQ242'] not in index"

Not sure if I did something wrong but found a workaround by not keeping it in features_to_keep so that the question to drop does not arise.

monindersingh:
Hi
Can you please provide some more detail? The code you used, the line where the error occurred, etc...best will be if you attach the code. I suspect this is an issue with the datafile but need to verify.

monindersingh:
Identified slight bug in code. Fixed and pull request created.

PR #60, opened 2018-12-10
Features to drop bug fix

monindersingh:
fix bug in handling of features_to_drop

### DIFF
--- a/aif360/datasets/standard_dataset.py
+++ b/aif360/datasets/standard_dataset.py
@@ -95,7 +95,8 @@ class StandardDataset(BinaryLabelDataset):
         if instance_weights_name:
             keep |= set([instance_weights_name])
         df = df[sorted(keep - set(features_to_drop), key=df.columns.get_loc)]
-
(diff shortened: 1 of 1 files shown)


adrinjalali:
Could you please add "Fixes #43" to the description, to link the PR to the issue?

It'd be also really nice if you could add a test which would fail on the old version, and works with this PR.

This way we can keep track of what exactly the PR is fixing.

monindersingh:
Fixes issue#43; previously, if a categorical feature was added to the features_to_drop list when defining a dataset i.e. it was a member of both the categorical_features and features_to_drop list, it was being dropped in the code from the overall set to features but not from the categorical_features list. Hence, there was an error subsequently when the categorical features were one-hot-encoded because this feature was still there in the categorical features list but no longer available in the data. Now, it is dropped from the categorical_features list as well.

hoffmansc:
Looks fine to me. I'm okay with no explicit test in this case but I agree in theory.

PR #64, opened 2019-01-15
added support for aif360.__version__

hoffmansc:
inspired by numpy

- change version from one central place in setup.py
- upon install, aif360/version.py is written which stores the installed version number
- makes version number accessible via:
```bash
> import aif360
> aif360.__version__
1.1.0
```


michaelhind:
Hi Sam,, it looks like it pass Python 3 test, but failed Python 2.  Was that expected?

PR #76, opened 2019-03-22
Added Area under Receiving Operating Characteristic curve 

connorhofenbi:
Hello,
       While working on a project for my group I noticed that IBM's AIF360 library did not have a way to measure area under the receiver operating characteristic curve. This is a metric I've found useful in the past and thought adding it to the library out of the box would be beneficial. 
        I was able to achieve this without importing any extra libraries. The "compute_ROC" function under metrics/utils.py takes in essentially the same parameters as "compute_num_TF_PN" . This function returns the false positive list, true positive list under different thresholds and area under the Receiver Operating Characteristic curve.
       Under metrics/classification_metric.py I added "compute_ROC_metric", which takes the privileged condition as a parameter and returns the same three variables as "compute_ROC".
       I tested this using the desperate impact remover example and received similar values to sklearn's library.
Thank you,
 Connor Hofenbitzer

### DIFF
--- a/aif360/metrics/utils.py
+++ b/aif360/metrics/utils.py
@@ -189,18 +189,16 @@ def compute_ROC(X, y_true, y_pred, w, feature_names, favorable_label,
     true_positive_matrix /= number_of_positives
     false_positive_maxtrix /= number_of_negatives
 
-    plt.title('Receiver Operating Characteristic')
-    plt.plot(false_positive_maxtrix, true_positive_matrix, 'b')
-    plt.legend(loc='lower right')
-    plt.plot([0, 1], [0, 1], 'r--')
-    plt.xlim([0, 1])
-    plt.ylim([0, 1])
-    plt.ylabel('True Positive Rate')
-    plt.xlabel('False Positive Rate')
+    #calculate AUC using trapezoidal sums
+    height = (np.add([true_positive_matrix[i] for i in range(1,true_positive_matrix.size)],
+                     [true_positive_matrix[x] for x in range(0,true_positive_matrix.size-1)]))
+    height /= 2.
+    width = -np.diff(false_positive_maxtrix)
 
-    plt.show()
+    #calculate AUC
+    auc = np.sum(height*width)
 
-    return false_positive_maxtrix, true_positive_matrix
+    return false_positive_maxtrix, true_positive_matrix, auc
 
 
 
--- a/aif360/metrics/utils.py
+++ b/aif360/metrics/utils.py
@@ -5,8 +5,6 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import numpy as np
-import matplotlib.pyplot as plt
-
 
 def compute_boolean_conditioning_vector(X, feature_names, condition=None):
     """Compute the boolean conditioning vector.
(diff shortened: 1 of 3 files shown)


animeshsingh:
@krvarshney can someone take a look at it?

connorhofenbi:
> Can we directly leverage `sklearn.metrics.roc_auc_score` function that takes `y_true` and `y_score` as inputs? That could make the code much simpler.

I could very easily, I was trying to reduce the amount of dependencies needed.

nrkarthikeyan:
I think it is ok to have sklearn dependency, since our long term plan is to take AIF360 closer to sklearn.

nrkarthikeyan:
Hi @connorhofenbi, wondering if you had a chance to think about my previous comment. Let us know your thoughts when you get a chance.

connorhofenbi:
> Hi @connorhofenbi, wondering if you had a chance to think about my previous comment. Let us know your thoughts when you get a chance.

My computer was broken for a but here's the change.

nrkarthikeyan:
Thanks @connorhofenbi . We retired support for python 2, so can you update the file accordingly (remove the `__future__` imports). Also, please set the favorable label to be positive when calling the ROC function.

aif360/metrics/utils.py
```diff
@@ -126,6 +120,29 @@ def compute_num_TF_PN(X, y_true, y_pred, w, feature_names, favorable_label,
         FN=np.sum(w[np.logical_and(y_true_pos, y_pred_neg)], dtype=np.float64)
     )
 
+def compute_ROC(y_true, scores, pos_label = 1, condition = None):
```
hoffmansc: you should use `favorable_label` here like above instead of `pos_label`.

also, `condition` should be implemented like above too (it's not currently used here)

aif360/metrics/utils.py
```diff
@@ -126,6 +120,29 @@ def compute_num_TF_PN(X, y_true, y_pred, w, feature_names, favorable_label,
...
+        condition (list(dict)): Same format as
+            :func:`compute_boolean_conditioning_vector`.
+
+    Returns:
+        False negative rate matrix for given classification thresholds (list)
+        True positive rate matrix for given classification thresholds  (list)
+        Area under Receiver Operating Characteristic Curve (float)
+    """
+
+    fpr, tpr, thresholds = metrics.roc_curve(y_true, scores, pos_label = pos_label)
```
hoffmansc: this should also take `instance_weights` (`sample_weight` in sklearn) into account.

it probably should use `roc_auc_score` but that doesn't seem to support the `pos_label` arg?

also, you're missing `sklearn` before `metrics`

aif360/metrics/utils.py
```diff
@@ -126,6 +120,29 @@ def compute_num_TF_PN(X, y_true, y_pred, w, feature_names, favorable_label,
...
+
+    #return values
+    return fpr, tpr, auc
```
hoffmansc: `auc` is never calculated

tests/test_classification_metric.py
```diff
@@ -112,4 +112,4 @@ def test_between_group():
         privileged_groups=[{'feat': 1}])
 
     b = np.array([0.5, 0.5, 1.25, 1.25, 1.25, 1.25])
-    assert cm.between_group_generalized_entropy_index() == 1/12*np.sum(b**2 - 1)
+    assert cm.between_group_generalized_entropy_index() == 1/12*np.sum(b**2 - 1)
```
hoffmansc: can you remove this so we don't have a superfluous commit?

aif360/metrics/utils.py
```diff
@@ -1,12 +1,6 @@
-"""This is the helper script for implementing metrics."""
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import print_function
-from __future__ import unicode_literals
-
+import sklearn
 import numpy as np
 
-
```
hoffmansc: you can leave 2 lines after the imports for consistency with the rest of the files

PR #79, opened 2019-04-29
hotfix for python 2.7 support in scipy

hoffmansc:
scipy 1.3.0 no longer supports Python 2.7

eventually we should do the same but for now we can do this

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ numpy>=1.14,<1.16
 matplotlib
 pandas==0.23.3
 pytest>=3.5.0
-scipy
+scipy<=1.2.1
 scikit-learn
 cvxpy>=1.0
 numba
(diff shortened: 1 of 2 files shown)


adrinjalali:
Any reason AIF360 still supports python2? It's a new library, it has no reason to support python2. This week probably the new sklearn will be released and it also doesn't support python2.

hoffmansc:
I agree. I can't think of a good reason to continue supporting it...

nrkarthikeyan:
Agree with @hoffmansc and @adrinjalali

hoffmansc:
Then maybe we should just do that now instead of this. Is there anything other than the README, Travis tests, and maybe setup.py that needs to be changed?

nrkarthikeyan:
@krvarshney, @michaelhind, @animeshsingh --- what do you all think about removing support for python 2? The one in pypi will still have support until the new version gets in there.

animeshsingh:
+1 on that. Python 2 should mostly be going out now...

michaelhind:
+1

krvarshney:
makes sense

krvarshney:
makes sense

hoffmansc:
Cool. Closing this then.

PR #114, opened 2019-10-03
Correction docstring

autoih:
close https://github.com/IBM/AIF360/issues/105


hoffmansc:
The way I understood it, `true labels` in this context meant ground-truth.

so ...false_positives should read: `weighted sum of predicted scores where true labels are 'unfavorable',` since scores are always for the favorable label

and then ...false_negatives, e.g., should be `weighted sum of 1 - predicted scores where true labels are 'favorable',`

and finally, ...true_negatives should be `weighted sum of 1 - predicted scores where true labels are 'unfavorable',`

@nrkarthikeyan do you agree? Should we rephrase it as `weighted sum of predicted 'scores' of instances with favorable ground-truth labels` or something?

autoih:
Thanks @hoffmansc. I've updated it.

PR #125, opened 2019-11-13
Differential fairness metric

hoffmansc:
Credit to Rashidul Islam for the reference implementation.

Adds `smoothed_empirical_differential_fairness()` to `BinaryLabelDatasetMetric`. This will use all protected attributes included in the dataset to compute the intersectional groups present. To ensure multi-class (non-binary) protected attributes are correctly considered, they should be mapped to integers in `custom_preprocessing`, e.g.:
```
def custom_preprocessing(df):
    df.race = df.race.map({'Black': 0, 'White': 1, 'Asian-Pac-Islander': 2,
                           'Amer-Indian-Eskimo': 3, 'Other': 4})
    return df
```
see examples in docs for more details.

### DIFF
--- a/aif360/metrics/binary_label_dataset_metric.py
+++ b/aif360/metrics/binary_label_dataset_metric.py
@@ -1,5 +1,3 @@
-from itertools import product
-
 import numpy as np
 from sklearn.neighbors import NearestNeighbors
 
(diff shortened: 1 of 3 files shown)


jrfoulds:
Thanks Sam, this is great!  Based on a quick look, the implementation looks correct to me and Rashid.  Our suggestions are mainly on the generality of the implementation.

-It may be useful to allow the user to set the concentration_parameter (which can be any non-negative number).  The current value can be the default.

-Differential fairness can also be applied in the multi-class setting, as well as the binary label setting.  If AIF360 doesn't usually consider this case, then it's fine to just have binary labels.  If you do want to add this in future, we could send you demo code for that as well.  The only difference is, for line 189 of binary_label_dataset_metric.py, instead of max(pos_ratio(i, j), neg_ratio(i, j)) it's a max over such ratios for all outcomes. 

-There is another version of our metric, differential fairness bias amplification, which calculates the difference in epsilon between the classifier and the data.  You may consider adding this to classification_metric.py.  In test_differential_fairness.py, this would correspond to calculating eps_clf - eps_data = 0.1066.

hoffmansc:
Thanks for the feedback!

I made `concentration` a parameter for the metric and I added the bias amplification metric. The changes I made should make it easier to implement multi-class down the road too but for now I think we should start with this.

jrfoulds:
Fantastic!  Thanks a lot, Sam and Prasanna!

dr97531:
That's great. Thanks, Sam and Prasanna.

PR #133, opened 2019-12-18
added predict_proba function to Adversarial Debiasing and added new example

guillemarsan:
- Added predict_proba function, which computes predicted scores, to Adversarial Debiasing by splitting predict function in two so Adversarial Debiasing can be coupled with post-processing algorithms such as Reject Option Classification for better performance.

- Added new example notebook demo_pre_in_postproc.ipynb as an example to show an algorithm of pre-processing (reweighing), in-processing (adversarial debiasing) and post-processing (reject option classification) working on the same dataset and their results.


hoffmansc:
Hi, @guillemarsan. Thanks for your PR.

Actually, in AIF360 we don't have `predict_proba()` functions. Instead, you can modify the `scores` attribute in addition to or instead of `labels` in `predict()`. In this case, I think it would be fine to just add a line:
```
new_dataset.scores = np.array(pred_labels, dtype=np.float64).reshape(-1, 1)
```

guillemarsan:
@hoffmansc, thank you for your message. I will change that and open a new PR.

PR #135, opened 2019-12-22
Added scores to Adversarial Debiasing and added new example

guillemarsan:
- Added  assignation to dataset scores in predict function of Adversarial Debiasing so it can be coupled with post-processing algorithms such as Reject Option Classification for better performance.

- Added new example notebook demo_pre_in_postproc.ipynb as an example to show an algorithm of pre-processing (reweighing), in-processing (adversarial debiasing) and post-processing (reject option classification) working on the same dataset and their results.


hoffmansc:
Hi, @guillemarsan. I'm good with the scores part but I don't think the example really adds enough new information to warrant another notebook.

guillemarsan:
Fair enough. I'll open another pull request only with the scores part.

PR #140, opened 2020-01-25
subset of dataset feature added

bhavyaghai:
Added the subset feature w.r.t. [issue 136](https://github.com/IBM/AIF360/issues/136)

### DIFF
--- a/aif360/datasets/structured_dataset.py
+++ b/aif360/datasets/structured_dataset.py
@@ -159,19 +159,24 @@ class StructuredDataset(Dataset):
             `StructuredDataset`: subset of dataset based on indexes
         """
 
+
         # convert each element of indexes to string
-        indexes_str = [str(x) for x in indexes]
+        indexes_str = list(map(str, indexes))
         # check if indexes is a subset of instance_names
         if not set(indexes_str).issubset(set(self.instance_names)):
             raise ValueError("indexes must be a subset of instance_names.")
 
+        pos = []
+        for i in indexes_str:
+            pos.append(self.instance_names.index(i)) 
+
         subset = self.copy()
         subset.instance_names = indexes_str
-        subset.features = self.features[indexes]
-        subset.labels = self.labels[indexes]
-        subset.instance_weights = self.instance_weights[indexes]
-        subset.protected_attributes = self.protected_attributes[indexes]
-        subset.scores = self.scores[indexes]
+        subset.features = self.features[pos]
+        subset.labels = self.labels[pos]
+        subset.instance_weights = self.instance_weights[pos]
+        subset.protected_attributes = self.protected_attributes[pos]
+        subset.scores = self.scores[pos]
         return subset
 
 
--- a/aif360/datasets/structured_dataset.py
+++ b/aif360/datasets/structured_dataset.py
@@ -151,32 +151,22 @@ class StructuredDataset(Dataset):
 
         
     def subset(self, indexes):
-        """ Subset of dataset based on instance names
+        """ Subset of dataset based on position
         Args:
-            indexes: iterable which contains instance names 
+            indexes: iterable which contains row indexes 
 
         Returns:
             `StructuredDataset`: subset of dataset based on indexes
         """
-
-
         # convert each element of indexes to string
         indexes_str = list(map(str, indexes))
-        # check if indexes is a subset of instance_names
-        if not set(indexes_str).issubset(set(self.instance_names)):
-            raise ValueError("indexes must be a subset of instance_names.")
-
-        pos = []
-        for i in indexes_str:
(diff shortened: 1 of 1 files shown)


aif360/datasets/structured_dataset.py
```diff
@@ -149,6 +149,32 @@ class description.
...
+
+        Returns:
+            `StructuredDataset`: subset of dataset based on indexes
+        """
+
+        # convert each element of indexes to string
+        indexes_str = [str(x) for x in indexes]
+        # check if indexes is a subset of instance_names
+        if not set(indexes_str).issubset(set(self.instance_names)):
+            raise ValueError("indexes must be a subset of instance_names.")
```
hoffmansc: This isn't actually necessary since the instance names aren't required to just be string-versions of integer values. They could, for example, be names of people. I think integer-indexing should be sufficient. Otherwise, looks good!

bhavyaghai: I also think that integer indexing should be good enough but I had to take this extra route because of the way instance_names have been defined in __init__ function. 

Please refer to structured_dataset.py  line 112
`
self.instance_names = df.index.astype(str).tolist()
`

Without this extra code, instance_names won't be consistent between the original dataset and its subset. instances_names for original dataset will be a list of strings while for its subset it will be list of integers.

Furthermore, this extra code also helps check for inconsistent index values like negative integers.   

Correction: this subset function is based on position and not on row names.

bhavyaghai: Please check the latest changes. Now, subset() works on row names rather than position.

For example, it can deal with the following case
'
ad = AdultDataset()
ad.instance_names[0] = 'i0'
ad.instance_names[4] = 'i4'
ad.subset(['i0', 'i4'])
'

bhavyaghai:
Please check the latest changes. Now, subset() works on row names rather than position.

For example, it can deal with the following case

ad = AdultDataset()
ad.instance_names[0] = 'i0'
ad.instance_names[4] = 'i4'
ad.subset(['i0', 'i4'])

hoffmansc:
@bhavyaghai, I think it would be better to keep using positional indexes instead of instance names. The `instance_names` should just be indexed with the same positions as the rest of the sample-wise arrays. We can just let numpy take care of errors for invalid indexes.

bhavyaghai:
@hoffmansc I think row names based subset function is a more generalized solution. If instance names are simple numbers, then both position-based and row names based techniques work fine. However, if row names are strings say a person's name, then position based subset will fail while row name based subset function will still work great!


Since, you like position based subset more, so I have rolled back to position based subset function. Please have a look at the latest changes

hoffmansc:
I think there's a breakdown in communication happening here. `instance_names` should not be set to `list(map(str, indexes))`, it should be indexed like the other fields (i.e. `[self.instance_names[i] for i in indexes]`) -- integer-location indexing.

I agree there could be some utility to name-based indexing. For example, Pandas has `loc` and `iloc` to differentiate the two. Since instance names are always strings in our case, you could combine both options in a single function, though, depending on if the input is a string-array or integer-array. This is up to you if you want to add that functionality.

Does this make sense?

bhavyaghai:
Quoting my previous comment
"I also think that integer indexing should be good enough but I had to take this extra route because of the way instance_names have been defined in init function.

Please refer to structured_dataset.py line 112
self.instance_names = df.index.astype(str).tolist()

Without this extra code i.e. `list(map(str, indexes))`, instance_names won't be consistent between the original dataset and its subset. instances_names for original dataset will be a list of strings while for its subset it will be list of integers."

If you have any counter to this argument, let me know. I will be happy to make required changes.

Secondly, I already implemented the functionality which can handle string array and integer array. Please refer to commit [subset function changed to row name based from index based](https://github.com/IBM/AIF360/pull/140/commits/76df8ee3ba93479de52504c55af7ffe971fe61df)

hoffmansc:
The line I suggested, `[self.instance_names[i] for i in indexes]`, does return a list of strings when `indexes` is an array of integers.

The commit you linked would not accommodate both `instance_names`-based _or_ integer-based indexing. It only accommodates the case where `indexes` is an array of types which can be cast as strings that match those in `instance_names`. If `instance_names` looks like:

```['John S.', 'Jane D.', 'Sam H.', ...]```

I would like to be able to either input, e.g., `subset([1, 2])` or `subset(['Jane D.', 'Sam H.'])`. The function should check the type of the input to determine if string or integer-based indexing is appropriate. If you think this will cause confusion then only integer position-based indexing should be allowed.

bhavyaghai:
I think I got the misunderstanding. Let's keep it simple and go with just integer-based indexing. More functionality can be added later.
On these lines, check the latest code and let me know if any more changes are required.

PR #155, opened 2020-03-05
Fix typo in mean_difference message

kcha:
This PR swaps the privileged and unprivileged terms in the messaging of `MetricTextExplainer.mean_difference()`.

If I understand correctly, the calculation for `mean_difference()` is unprivileged instances - privileged instances as seen here: https://github.com/IBM/AIF360/blob/9304749698a50ff3c2118619c627e49df626ecf3/aif360/metrics/dataset_metric.py#L77


hoffmansc:
Good catch!

PR #157, opened 2020-03-09
Added explanation of sep argument to convert_to_dataframe() in  struc…

jimbudarz:
…tured_dataset.py


jimbudarz:
The purpose of this PR is to clarify the usage of convert_to_dataframe() in the documentation, as Pandas' get_dummies has a different default ('_') prefix value from this package ('=')
https://aif360.readthedocs.io/en/latest/modules/datasets.html

PR #159, opened 2020-03-10
Clarification that reweighing algorithm only modifies the instance_weights attribute

jimbudarz:


jimbudarz:
There was no other place that mentioned the name of the attribute in the Reweighing algorithm's documentation- you'd otherwise need to check for potentially relevant attributes in the Datasets documentation.

PR #165, opened 2020-04-02
Muliti class support

swapna-somineni:
This PR has the following changes

Added new class **MulticlassLabelDataset** to support the multiple values in the favorable and unfavorable label's. To convert the multiple favorable and unfavorable values into binary labels, we are iterating through the labels values in the dataframe and check whether the label value is  in the favorable list, if it is true then assign the favorable label value as 1 otherwise 0 for unfavourable values.

### DIFF
--- a/aif360/datasets/multiclass_label_dataset.py
+++ b/aif360/datasets/multiclass_label_dataset.py
@@ -1,0 +1,69 @@
+import numpy as np
+
+from aif360.datasets import StructuredDataset
+'''
+Multiclass supports the multiple values in the favorable and unfavorable label's
+'''
+class MulticlassLabelDataset(StructuredDataset):
+    """Base class for all structured datasets with multiclass labels."""
+
+    def __init__(self, favorable_label = [1.], unfavorable_label = [0.], **kwargs):
+        """
+        Args:
+            favorable_label (list): Label value which is considered favorable
+                (i.e. "positive").
+            unfavorable_label (list): Label value which is considered
+                unfavorable (i.e. "negative").
+            **kwargs: StructuredDataset arguments.
+        """
+        self.favorable_label = favorable_label
+        self.unfavorable_label = unfavorable_label
+
+        super(MulticlassLabelDataset, self).__init__(**kwargs)
+
+        fav_label_value = 1.
+        unfav_label_value = 0.
+        #Iterate the labels values in the dataframe and check whether the label value is 
+        # in the favorable list, if it is true then assign the favorable label value as 1
+        # otherwise 0 for unfavourable values
+        for index in range(0,len(self.labels)):
+            if self.labels[index] in self.favorable_label:
+                self.labels[index] = float(fav_label_value)
+            elif self.labels[index] in self.unfavorable_label:
+                self.labels[index] = float(unfav_label_value)
+        
+        self.favorable_label = float(fav_label_value)
+        self.unfavorable_label = float(unfav_label_value)
+
+       
+
+    def validate_dataset(self):
+        """Error checking and type validation.
+​
+        Raises:
+            ValueError: `labels` must be shape [n, 1].
+            ValueError: `favorable_label` and `unfavorable_label` must be the
+                only values present in `labels`.
+        """
+        # fix scores before validating
+        if np.all(self.scores == self.labels):
+            for i in range(0,len(self.scores)):
+                if self.scores[i] in self.favorable_label:
+                    self.scores[i] = float(1)
+                else:
+                    self.scores[i] = float(0)
+
+        super(MulticlassLabelDataset, self).validate_dataset()
+
+        # =========================== SHAPE CHECKING ===========================
+        # Verify if the labels are only 1 column
+        if self.labels.shape[1] != 1:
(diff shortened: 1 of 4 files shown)


aif360/datasets/multiclass_label_dataset.py
```diff
@@ -0,0 +1,69 @@
...
+        self.unfavorable_label = unfavorable_label
+
+        super(MulticlassLabelDataset, self).__init__(**kwargs)
+
+        fav_label_value = 1.
+        unfav_label_value = 0.
+        #Iterate the labels values in the dataframe and check whether the label value is 
+        # in the favorable list, if it is true then assign the favorable label value as 1
+        # otherwise 0 for unfavourable values
+        for index in range(0,len(self.labels)):
```
hoffmansc, line 29: Could we move this remapping to happen in the `__init__()` function of binary_label_dataset_metric.py and classification_metric.py? That way, the `MulticlassLabelDataset` class can keep the original labels and stay general for the future.

swapna-somineni: @hoffmansc As per the inputs from the last meeting, the code was moved from classification_metric.py to the `MulticlassLabelDataset`, so that the changes related to the multi class will be in single class.

hoffmansc: Right. I don't want to change the implementations in classification_metric.py, just handle the binarization of the labels there. Otherwise, we lose the information in `MultiClassDataset` labels. Right now, it's effectively no different than a `BinaryLabelDataset` as far as I can tell

swapna-somineni: @hoffmansc Updated the code as per the suggestion. Please review. Thanks!

aif360/metrics/binary_label_dataset_metric.py
```diff
@@ -26,13 +27,28 @@ def __init__(self, dataset, unprivileged_groups=None, privileged_groups=None):
...
+            fav_label_value = 1.
+            unfav_label_value = 0.
+            #Iterate the labels values in the dataframe and check whether the label value is 
+            # in the favorable list, if it is true then assign the favorable label value as 1
+            # otherwise 0 for unfavourable values
+            for index in range(0,len(self.dataset.labels)):
+                if self.dataset.labels[index] in self.dataset.favorable_label:
+                    self.dataset.labels[index] = float(fav_label_value)
+                elif self.dataset.labels[index] in self.dataset.unfavorable_label:
+                    self.dataset.labels[index] = float(unfav_label_value)
```
hoffmansc, line 48: This will modify the original dataset which we don't want to do. We should make a copy first and then change the labels, i.e.:
```py
self.dataset = self.dataset.copy()
# find all labels which match any of the favorable labels
fav_idx = np.logical_or.reduce(np.equal.outer(self.dataset.labels, self.dataset.favorable_label))
# replace labels with corresponding values
self.dataset.labels = np.where(fav_idx, fav_label_value, unfav_label_value)
```

aif360/metrics/classification_metric.py
```diff
@@ -32,20 +33,35 @@ def __init__(self, dataset, classified_dataset,
...
+            fav_label_value = 1.
+            unfav_label_value = 0.
+            #Iterate the labels values in the dataframe and check whether the label value is 
+            # in the favorable list, if it is true then assign the favorable label value as 1
+            # otherwise 0 for unfavourable values
+            for index in range(0,len(self.classified_dataset.labels)):
+                if self.classified_dataset.labels[index] in self.classified_dataset.favorable_label:
+                    self.classified_dataset.labels[index] = float(fav_label_value)
+                elif self.classified_dataset.labels[index] in self.classified_dataset.unfavorable_label:
+                    self.classified_dataset.labels[index] = float(unfav_label_value)
```
hoffmansc, line 60: same as above

tests/test_classification_metric.py
```diff
@@ -108,3 +109,82 @@ def test_between_group():
...
+    unfavorable_values = [2]
+    bld = MulticlassLabelDataset(favorable_label = favorable_values, unfavorable_label = unfavorable_values , df = df , label_names=['label'],
+        protected_attribute_names=['feat'])
+    bld2 = MulticlassLabelDataset(favorable_label = favorable_values, unfavorable_label = unfavorable_values , df=df2, label_names=['label'],
+        protected_attribute_names=['feat'])
+    cm = ClassificationMetric(bld, bld2, unprivileged_groups=[{'feat': 2}],
+        privileged_groups=[{'feat': 0},{'feat': 1}])
+    confusion_matrix = cm.binary_confusion_matrix()
+    assert confusion_matrix == {'TP': 7.0, 'FN': 1.0, 'TN': 2.0, 'FP': 0.0}
+
```
hoffmansc, line 141: can you also add an assert to make sure the original labels are preserved by `ClassificationMetric`?

aif360/metrics/binary_label_dataset_metric.py
swapna-somineni: Thanks Sam. Updated the code as per the suggestion.

PR #174, opened 2020-05-06
Refactor make_scorer for group fairness metrics

chkoar:


aif360/sklearn/metrics/metrics.py
```diff
@@ -119,42 +119,34 @@ def ratio(func, y, *args, prot_attr=None, priv_group=1, sample_weight=None,
     return numerator / denominator
 
 
-# =========================== SCORER FACTORIES =================================
-def make_difference_scorer(diff_func):
-    """Make a scorer from a 'difference' metric (e.g.
+# =========================== SCORER FACTORY =================================
+def make_scorer(score_func, ratio, **kwargs):
```
hoffmansc, line 123: Can we change `ratio` to `is_ratio` and set the default to False? Difference metrics are more common.

tests/sklearn/test_metrics.py
```diff
@@ -1,5 +1,7 @@
 import numpy as np
+import pytest
 from sklearn.linear_model import LogisticRegression
+from sklearn.utils.testing import assert_almost_equal
```
hoffmansc, line 4: Can we change this to `from numpy.testing import assert_almost_equal`? That's where sklearn get it and I was having trouble finding the documentation at first.

tests/sklearn/test_metrics.py
```diff
@@ -101,3 +103,18 @@ def test_between_group_generalized_entropy_index():
...
+    "func, is_ratio",
+    [
+        (statistical_parity_difference, False),
+        (disparate_impact_ratio, True),
+        (equal_opportunity_difference, False),
+        (average_odds_difference, False),
+    ],
+)
+def test_make_scorer(func, is_ratio):
+    actual = func(y, y_pred, prot_attr="sex")
```
hoffmansc, line 117: The test should also make sure if the `abs()` or `min()` condition applies. Maybe you could flip the protected attributes and make sure it's still equal?

chkoar:
Hey @hoffmansc, I think that addressed your comments. Is there something else? 

Apart from that I think that the scorers could be patched (probably using a patch function) in the default scorers of `scikit-learn` so one could run:
```python
cross_val_score(model, X, y, scoring="statistical_parity_difference")
```
or getting the scorer from `sklearn.metrics.get_scorer` just passing the string:

```python
from sklearn.metrics import get_scorer
scorer = get_scorer("statistical_parity_difference")
```

Monkey-patching might be a controversial thing in the community. 
If this is not the case, I could add this functionality in this or another PR.

hoffmansc:
@chkoar I don't feel that strongly about it but I do think it might add a little bit of confusion about what functionality coming from which library... For now, the current syntax doesn't seem too cumbersome to me

chkoar:
> but I do think it might add a little bit of confusion about what functionality is coming from which library

This might be true. In any case I opened an [issue](https://github.com/scikit-learn/scikit-learn/issues/17168) in the `scikit-learn`'s upstream repo.

PR #183, opened 2020-06-10
add score attribute to fair data set

baba-mpe:
Prediction scores are computed as part of Zemel's lfr algorithm. However, so far the `dataset_new` class does not allow for outputting those scores, but only for the resulting binary predicted labels. In this minor change we added a `dataset_new.scores` attribute which enables scores output. This is analogous to already existing code, see, e.g., in the adversarial debiasing code: https://github.com/IBM/AIF360/blob/master/aif360/algorithms/inprocessing/adversarial_debiasing.py#L260


hoffmansc:
Duplicate of #184

PR #190, opened 2020-07-08
Update readme and maintainers list

SSaishruthi:
- This PR updates the maintainer list with R maintainers. Arranged in alphabetical order based on the last name.
- Adding about R version to the readme (moving it to the first paragraph so that it will have visibility)

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,14 @@ Get in touch with us on [Slack](https://aif360.slack.com) (invitation
 
 ## Setup
 
+### R
+
+``` r
+install.packages("aif360")
+```
+
+For more details regarding the R setup, please refer to instructions [here](aif360/aif360-r/README.md).
+
 ### Python
 
 Supported Python Configurations:
@@ -67,14 +75,6 @@ Supported Python Configurations:
 | macOS   | 3.5, 3.6, 3.7  |
 | Ubuntu  | 3.5, 3.6, 3.7  |
 | Windows | 3.5, 3.6, 3.7  |
-
-### R
-
-``` r
-install.packages("aif360")
-```
-
-For more details regarding the R setup, please refer to instructions [here](aif360/aif360-r/README.md).
 
 ### (Optional) Create a virtual environment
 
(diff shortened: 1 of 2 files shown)


SSaishruthi:
@hoffmansc @gdequeiroz 
Can we please take a look at it before migration?

gdequeiroz:
@SSaishruthi It looks good. Can you also add/change: 

- the CRAN badge ([CRAN\_Status\_Badge](https://cran.r-project.org/package=aif360)) beside the pypi package.

image

- Remove the word Python from here: "The AI Fairness 360 **Python** package includes..." -> "The AI Fairness 360 package includes"

- In the Setup section (https://github.com/IBM/AIF360#setup), I would do something like this:

image

SSaishruthi:
@gdequeiroz Updated as per your comments

README.md
```diff
@@ -56,7 +58,7 @@ Get in touch with us on [Slack](https://aif360.slack.com) (invitation
 
 ## Setup
 
-AIF360 is available for both Python and R. For R setup, please refer to instructions [here](aif360/aif360-r/README.md).
+### Python
```
gdequeiroz, line 61: I would add R first and then Python since the instructions that follow are for Python

SSaishruthi: @gdequeiroz I have removed the line from `setup` and added under first paragraph. Arranged in alphabetical order.

SSaishruthi: I did that because users will know when they start reading through the readme that it is available in both versions

README.md
```diff
@@ -3,10 +3,12 @@
 [![Build Status](https://travis-ci.org/IBM/AIF360.svg?branch=master)](https://travis-ci.org/IBM/AIF360)
 [![Documentation](https://readthedocs.org/projects/aif360/badge/?version=latest)](http://aif360.readthedocs.io/en/latest/?badge=latest)
 [![PyPI version](https://badge.fury.io/py/aif360.svg)](https://badge.fury.io/py/aif360)
+[![CRAN\_Status\_Badge](http://www.r-pkg.org/badges/version/aif360)](https://cran.r-project.org/package=aif360)
 
 The AI Fairness 360 toolkit is an extensible open-source library containg techniques developed by the
-research community to help detect and mitigate bias in machine learning models throughout the AI application lifecycle.
-The AI Fairness 360 Python package includes
+research community to help detect and mitigate bias in machine learning models throughout the AI application lifecycle. AI Fairness 360 package is available in both Python and R.
```
SSaishruthi, line 9: @gdequeiroz added here

README.md
hoffmansc: I agree with @gdequeiroz, @SSaishruthi can we move R instructions above Python? It's weird to me to have Python, then R, then back to Python instructions.

PR #195, opened 2020-07-29
Reconcile code and docstring for consistency_score and consistency

leenamurgai:
The documentation for the functions consistency_score and consistency
match the paper, which has a factor of 1 over k (k=n_neighbors in the
AIF360 documentation) to the left of the outer sum, rather than to the
left of the inner sum. The code in both functions does not match the
documentation since it takes the mean of y for the k nearest neighbours
(rather than the sum) and doesn't divide the outer sum by n_neighbors.
    
To see the error in the previous documentation, consider the extreme
case where y_hat = 1 for all examples. Here the consistency_score should
be 1 as implemented in the code.

### DIFF
--- a/aif360/metrics/binary_label_dataset_metric.py
+++ b/aif360/metrics/binary_label_dataset_metric.py
@@ -126,8 +126,8 @@ class BinaryLabelDatasetMetric(DatasetMetric):
         labels are for similar instances.
 
         .. math::
-           1 - \frac{1}{n\cdot\text{n_neighbors}}\sum_{i=1}^n |\hat{y}_i -
-           \sum_{j\in\mathcal{N}_{\text{n_neighbors}}(x_i)} \hat{y}_j|
+           1 - \frac{1}{n}\sum_{i=1}^n |\hat{y}_i -
+           \frac{1}{\text{n_neighbors}} \sum_{j\in\mathcal{N}_{\text{n_neighbors}}(x_i)} \hat{y}_j|
 
         Args:
             n_neighbors (int, optional): Number of neighbors for the knn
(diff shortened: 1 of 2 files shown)


hoffmansc:
I think you're right but in this case k is n_neighbors in our documentation so that should be deleted from the first denominator and swapped for k.

Could you also do this for the `BinaryLabelDatasetMetric` version?

leenamurgai:
Thanks! I thought replacing k with n_neighbors was the best option since it was the minimal change required to make the code and documentation consistent. n_neighbours though a little verbose is also more descriptive than just k I suppose. I have implemented the same change in the BinaryLabelDatasetMetric version as requested also - thanks for pointing this out.

aif360/metrics/binary_label_dataset_metric.py
```diff
@@ -127,7 +127,7 @@ def consistency(self, n_neighbors=5):
 
         .. math::
            1 - \frac{1}{n\cdot\text{n_neighbors}}\sum_{i=1}^n |\hat{y}_i -
```
hoffmansc, line 129: This line is what I was referring to. It has an extra n_neighbors now that should be removed.

Also, if you could put the `n_neighbors` in \text mode that will make it render nicer.

leenamurgai: Ah yes, okay. Done, hopefully it works. In LaTeX you need to escape underscores but not here it seems.

leenamurgai:
Is this all good now?

leenamurgai:
Are we good to merge?

FYI, Richard Zemel has now also corrected the paper now and updated it on his website: http://www.cs.toronto.edu/~zemel/documents/fair-icml-final.pdf.

hoffmansc:
> @hoffmansc, are we good to merge?

Nothing more needed from your end. Just waiting for #196 to be approved which should fix the build issue. Sorry for the delay...

leenamurgai:
Good to know. Thanks.

hoffmansc (approved):

PR #214, opened 2020-11-11
Add generic benefit function for inequality indices

leenamurgai:
Inequality indices measure how unevenly distributed the benefits of an algorithm are distributed.  In the case of a binary classifier, all algorithmic predictions (where the ground truth is known) can be categorised in a confusion matrix as either a true positive, false positive, false negative or true negative. A benefit_function can then be defined by simply assigning a value to each of these four cases. I have given the user the ability to choose this mapping.

The code now uses the confusion matrix to calculate the benefits rather than the predictions (y_pred) and ground truth (y_true). The code now consistently uses the sample_weights specified in the dataset class in calculating the benefits.

I have refactored the code to separate the calculation of the index into two parts, first calculating the benefits array and then calculating the index from the benefits. This reduces repetition of code. Note that the order of the benefits does not matter in the calculation of the indices.

I have added tests that both do not specify the benefit function and specify the benefit function as the default value. I have further added tests for two additional benefit_function choices (both from the original paper), one of them being the case where not all data points result in a benefit value. There are two tests which now fail, test_generalized_entropy_index and test_between_group_generalized_entropy_index in test_metrics. These tests are now obsolete since the sklearn compatible API does not account for sample_weights. I have commented the tests out as I am not sure why the code in aif360 sklearn.metrics does not simply call the code from aif360.metrics to avoid duplication.


leenamurgai:
Are there any thoughts on this?

krvarshney:
This is a good suggestion. The benefit function in the original paper by Speicher et al. might not be the best in many cases: taking 2 for false favorable labels (false positives), 1 for true favorable labels (true positives), 1 for true unfavorable labels (true negatives), and 0 for false unfavorable labels (false negatives). This recommendation is seemingly consistent with the “what you see is what you get” worldview because it is examining model performance, assumes the costs of false positives and false negatives are the same, and takes the perspective of affected users who want to be allocated the favorable outcome even if they are not truly suitable candidates. More appropriate benefit functions for many problem specifications may be benefit functions that are either (1) 1 for true favorable and true unfavorable labels and 0 for false favorable and false unfavorable labels (“what you see is what you get” while balancing societal needs), or (2) 1 for true favorable and false favorable labels and 0 for true unfavorable and false unfavorable labels (“we’re all equal”).

leenamurgai:
I've done some work looking at generalised entropy indices and concluded they are not measures of fairness at all, but rather measures of utility. In particular you can rewrite these metrics as a function of two parameters (model accuracy and mean error). Here's my proof:

GEI-Utlity- proof

In addition, it can be shown that when alpha=0 minimising the GEI is equivalent to minimising the cross entropy loss. When alpha=2 you can instead take the error as your benefit (rather than one plus the error since negative benefits are permitted). In this case minimising the GEI is equivalent to minimising the mean squared error.

Given this, I do believe it is misleading to have these indices categorised as individual fairness metrics alongside consistency (https://aif360.readthedocs.io/en/latest/modules/sklearn.html?highlight=individual#individual-fairness-metrics). Individual fairness as defined by Dwork et. al. in Fairness Through Awareness 2011 (approximated by the metric consistency) is importantly _not_ a measure of utility.

Finally, an important consequence of this is that the trade-off (empirically evidenced in the paper which is referenced here: https://aif360.readthedocs.io/en/latest/modules/generated/aif360.sklearn.metrics.generalized_entropy_error.html#aif360.sklearn.metrics.generalized_entropy_error) is not between group and individual fairness but rather between utility and group fairness metrics.

hoffmansc:
@leenamurgai thanks for your analysis on this. Your proof looks good but I have some questions about the conclusions:
- you mention when $\alpha=2$ minimizing GEE is equivalent to minimizing MSE -- can you provide your proof for that? According to proposition 3.2 from the paper, this should only be true if the data is fully separable. In general, if you could explain how your conclusion relates to proposition 3.2 I think that would be super helpful because from what I can tell they seem to contradict.
- how do you define a measure of utility in this context? I think the key point is that $\lambda$ and $\mu$ measure different "utility" functions and the interplay between them is what the index measures. Additionally, I think it's important to point out that entropy indices measure a different definition of individual fairness than that of "Fairness through Awareness." In fact, GEI has no dependence on X at all (what the authors call "anonymity") only y and y_pred. In that sense, it appears similar to traditional utility functions but I would argue it can still be a measure of fairness.

leenamurgai:
@hoffmansc Thank you for your response. Regarding your question on $\alpha=2$, I should have said that $I_2(\boldsymbol{b})$ is a linear function of the mean squared error. Here's my proof:

$$
\begin{align}
I_2(\boldsymbol{b}) & = \frac{1}{2n} \sum_{i=1}^n (x_i^2-1) \\
& = \frac{1}{2n\mu^2} \sum_{i=1}^n [ ( \hat{y_i}-y_i )^2 + 2( \hat{y_i}-y_i+1 ) - 1 - \mu^2 ] \\
& = \frac{1}{2n\mu^2} \sum_{i=1}^n (\hat{y_i}-y_i)^2 + \frac{1}{\mu} - \frac{1}{2\mu^2} - \frac{1}{2}
\end{align}
$$

The Index is equivalent to a cost sensitive utility, where the costs in the matrix are not fixed but depend on $\mu$ and $\alpha$. The latter is fixed in advance of training but the former is not. You might argue that this is not exactly utility since the cost matrix does not obey the rules of a _reasonable_ cost matrix (in that the cost of an error at every point in $(\lambda,\mu)$ space is not always greater than the cost of an accurate prediction). Nevertheless, the behaviour does not differ significantly from utility specifically in the case $\alpha=2$.

As mentioned, it is the case that in part of the model solution space $(\lambda, \mu)$ one can reduce the index by decreasing the accuracy, but for $\alpha=2$, this only happens when $\mu>1.667$ (there are lots of false positives). This can only happen if the model is sufficiently inaccurate because $\lambda\leq\mu\leq2-\lambda$. To see this note that, $\mu$ is minimal when all the errors are false negatives and maximal when all the errors are false positives. For any reasonable model $0.5\leq\mu\leq1.5$ since the accuracy $0.5\leq\lambda\leq1$. If we fix the dataset, we know $\mathbb{P}(Y=1)$ and so our accuracy bounds are tightened and so too the bounds for $\mu$. For fixed $\mu$, the index is a monotonically decreasing function of accuracy. Here are some plots showing the behaviour of the $I_2(\mu)$:

| $I_2(\mu)$ for fixed values of $\lambda$ | Contour plot for $I_2(\mu, \lambda)$ |
|-|-|
| Fig_I_alpha_2_varlambda | Fig_ccplot_alpha2 |

In the contour plot I mark off the area where exchanging an accurate prediction for an erroneous (false positive) prediction reduces the value of the index. The area is small and importantly doesn't exist once the accuracy exceeds 66.7%. After this, the only way to reduce the index value is by increasing the accuracy of the model.

In proposition 3.2, the Bayes optimal model is the least accurate any _reasonable_ model should be. $\hat{y}=\theta^A(x)=0$ which has an accuracy of $\lambda=1-p=\mathbb{P}(Y=1)$ where $0 < p < 0.5$. The less accurate model they fabricate in order to reduce the value of the index $\hat{y}=\theta^q(x)$, is not _reasonable_, it has an accuracy of $2p(1-p)<0.5$.

While you might still feel that generalised entropy indices provide some measure of fairness, I don't feel it can be argued that they provide a measure of _individual fairness_. The definitions of _individual fairness_ according to [Dwork et. al.](https://arxiv.org/abs/1104.3913) vs [Speicher, Heidari, et. al](https://dl.acm.org/doi/10.1145/3219819.3220046) are in fact contradictory. According to [Dwork et. al.](https://arxiv.org/abs/1104.3913) _individual fairness_ is a property of a single map from $x$ to $y$ or $x$ to $\hat{y}$. It is not the difference between two maps, $\hat{y}(x)$ and $y(x)$. Although [Speicher, Heidari, et. al](https://dl.acm.org/doi/10.1145/3219819.3220046) claim their definition is in line with Dwork et. al., as far as I can understand, it is not.

I have more complete analysis of generalised entropy indices in [chapter 5 of my book Mitigating Bias in Machine Learning](https://mitigatingbias.ml/#ch_InequalityIndices) and would be grateful for your feedback, especially if you disagree.

hoffmansc:
@leenamurgai thanks for your write-up and for linking your book -- it's very clear. This is very interesting! I understand where you're coming from here and I agree it's misleading to imply entropy indices measure the Dwork, et al. individual fairness definition. Perhaps we could categorize them separately as, say, "distributional fairness"? I'd also like to include a brief summary/disclaimer of your findings in the documentation as well as maybe a notebook just showing the graphs and an example and linking to your chapter. What's your recommendation for how we should present these?

In your opinion, is there any merit to calculating the _overall_ generalized entropy index or only between-group? What about within-group?

hoffmansc:
Regarding your PR itself, I have a couple suggestions:

1. Can we just name the cases from the table (e.g., "accuracy", "luck") and translate those to weights? If someone really wants something else, maybe then they could provide a dict or just the raw benefits? I'm also wondering about extensions to regression or multiclass tasks. I think this is why we have a more generic `generalized_entropy_index` function in the sklearn API which accepts benefit values directly.
2. Can we have the function handle n/a (`None`) weights in the benefit function? e.g., if we want FPR equality. This is not supported yet, right?

leenamurgai:
Just thinking about this, there are two separate issues here, one is the original pull request which added some flexibility and efficiency in the code and another is the categorisation of the metrics. Would it be a good idea to separate these? How would be best? A separate pull request / issue on the documentation for the categorisation?

PR #215, opened 2020-11-11
Fairlearn wrappers

sohiniu:
This PR includes Exponentiated Gradient Reduction and Grid Search Reduction wrapped from the fairlearn library


myampol:
Mmk okc to
________________________________

myampol:
Please disregard - the below "message" resulted from an errant keyboard, and is without meaning.

Apologies.

________________________________

nrkarthikeyan (approved):

nrkarthikeyan:
@hoffmansc can  you do a quick check?

aif360/sklearn/inprocessing/exponentiated_gradient_reduction.py
```diff
@@ -0,0 +1,143 @@
...
+        Returns:
+            numpy.ndarray: Predicted class label per sample.
+        """
+        if self.drop_prot_attr:
+            X = X.drop(self.prot_attr_cols, axis=1)
+        
+        return self.classes_[self.model.predict(X)]
+
+
+    def predict_proba(self, X):
```
MiroDudik, line 127: Glad to see you're incorporating some algorithms from Fairlearn! But please note that `_pmf_predict` is not the same as `predict_proba`: https://github.com/fairlearn/fairlearn/issues/5.

dennislwei: @MiroDudik thanks for noticing and for the comment. I understand that Fairlearn returns randomized classifiers and so one should probably 1) sample a classifier from the collection, with probability given by its weight, and 2) call `predict_proba` on this classifier. This would be similar to the behavior of `predict` [here](https://github.com/fairlearn/fairlearn/blob/master/fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py#L201). Would this be your recommendation? If so, do you have a function or method that does this? Perhaps your colleague Roman Lutz can weigh in as well since he's active on the issues you linked: https://github.com/fairlearn/fairlearn/issues/5, https://github.com/fairlearn/fairlearn/issues/65.

MiroDudik: Your suggested fix is correct--and definitely something I'd recommend (and also something we should add to exponentiated gradient in Fairlearn). It's also exactly what we do in Fairlearn when exponentiated gradient is run on a regression problem (in that case _pmf_predict has a different return type). Tagging @romanlutz for awareness.

nrkarthikeyan: @dennislwei, @sohiniu - should we  incorporate this now or track it using a  bug and  do this later?

dennislwei: Here's some code (mostly taken from [Fairlearn](https://github.com/fairlearn/fairlearn/blob/master/fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py)) that should be not too far off.

```
    from sklearn.utils import check_random_state
    from sklearn.utils.validation import check_is_fitted

    check_is_fitted(self.model)
    random_state = check_random_state(random_state)

    pred = pd.DataFrame()
    for t in range(len(self.model._hs)):
        if self.model.weights_[t] == 0:
            pred[t] = np.zeros(len(X))
        else:
            pred[t] = self.model._hs[t](X)

    randomized_pred = np.zeros(pred.shape[0])
    for i in range(pred.shape[0]):
        randomized_pred[i] = random_state.choice(pred.iloc[i, :], p=self.model.weights_)
    return np.concatenate((1-randomized_pred, randomized_pred), axis=1)
```

@MiroDudik @romanlutz I don't know whether `_hs` is the right method/function to call. Ideally the predictors in the `predictors_` attribute would have a `predict_proba` method that we could call, but I don't see `predictors_` being used.

nrkarthikeyan:
Opened #226 and #225 to track the issues mentioned here.

Issue #152, opened 2020-02-23
Port in-processing algorithms to sklearn-compatible API


hoffmansc:
- [X] AdversarialDebiasing
- [ ] ARTClassifier
- [ ] MetaFairClassifier
- [ ] PrejudiceRemover

PR #254, opened 2021-07-20
Increasing max_iter to 1000 for LogisticRegression used in PrejudiceRemover

mfeffer:
The `PrejudiceRemover` inprocessing bias mitigation algorithm uses `LogisticRegression`, and by default, `max_iter` for `LogisticRegression` from sklearn is 100. However, in practice, more iterations are occasionally needed to fit `PrejudiceRemover` to large datasets, some of which are common benchmarks in the fairness literature.

With this in mind, this PR increases the `max_iter` threshold to 1000 from 100 in order to help `PrejudiceRemover` properly converge on large datasets.


nrkarthikeyan:
@mfeffer - please verify the DCO error and sign off.

Also, can we make this a user-configurable option, so that the default is 100 but can be changed to 1000 or more by users who want them? It will make it easier for users who have smaller datasets to run things quickly.

hoffmansc (approved):
I'm fine with this for now. max_iter is just a budget -- it may use less if possible.

Making this user-configurable will probably have to wait until we rewrite the code for the sklearn-compatible port (#152)

PR #259, opened 2021-08-11
Update component.yaml to kfp v2 sdk

yhwang:
Update component.yaml to kfp v2 compatible. In v2,
you need to declare the data type for all of the input/output
arguments.

Signed-off-by: Yihong Wang <yh.wang@ibm.com>

### DIFF
--- a/mlops/kubeflow/bias_detector_pytorch/component.yaml
+++ b/mlops/kubeflow/bias_detector_pytorch/component.yaml
@@ -15,20 +15,20 @@ description: |
 metadata:
   annotations: {platform: 'OpenSource'}
 inputs:
-  - {name: model_id,                     description: 'Required. Training model ID', default: 'training-dummy'}
-  - {name: model_class_file,             description: 'Required. pytorch model class file', default: 'PyTorchModel.py'}
-  - {name: model_class_name,             description: 'Required. pytorch model class name', default: 'PyTorchModel'}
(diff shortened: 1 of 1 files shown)


Tomcli:
So KFP is working on a new v2 design that requires few extra fields from its components. Part of the requirements is to specify the input/output parameter type for all the KFP components for the new design. This PR is to address this change.

nrkarthikeyan:
@Tomcli , @yhwang - does this change preserve backward compatibility?

Tomcli:
> @Tomcli , @yhwang - does this change preserve backward compatibility?

No, the `type` field was optional in the old KFP version. However, the upcoming v2 KFP version will make the `type` required sometimes next year, so we want to update the AIF360 component to be ready for the new KFP versions.

nrkarthikeyan:
From your response, it seems like `type` field is optional now, but will become mandatory later. So including the `type` field would preserve backward compatibility right? Just making sure it wont break any existing code.

Tomcli:
> From your response, it seems like `type` field is optional now, but will become mandatory later. So including the `type` field would preserve backward compatibility right? Just making sure it wont break any existing code.

yes it will preserve backward compatibility. Nothing will be break from this PR. Sorry I probably didn't interpret your message correctly in my last comment.

nrkarthikeyan (approved):

nrkarthikeyan:
Ok approved and merged.

Tomcli:
thanks @nrkarthikeyan

PR #260, opened 2021-08-13
[WIP] Add regression metric

SSaishruthi:
This PR adds regression metric to the toolkit

### DIFF
--- a/aif360/metrics/regression_metric.py
+++ b/aif360/metrics/regression_metric.py
@@ -1,0 +1,12 @@
+from sklearn.metrics import mean_absolute_error
+import numpy as np
+
+def mae(dataset, dataset_pred):
+    # Get the sensitive attribute
+    sensitive_attribute = dataset.protected_attribute_names[0]
+    # Get index of the first sensitive attribute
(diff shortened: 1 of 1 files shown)


SSaishruthi:
@nrkarthikeyan This is a basic skeleton I was thinking. We have a bunch of methods to group datasets and get indexes. I did not use existing methods. Rather created a function explaining the functionality. Can you please take a look and provide suggestions?

nrkarthikeyan:
@SSaishruthi - can you verify the DCO thing and sign off?

nrkarthikeyan:
Can we do this directly on the sklearn version? In that case we can follow the pattern in https://github.com/Trusted-AI/AIF360/blob/master/aif360/sklearn/metrics/metrics.py

Is there any reason why the classical version is needed?

nrkarthikeyan:
One place to start if you want to use the "classic" version is to start with the regression dataset class https://github.com/Trusted-AI/AIF360/blob/master/aif360/datasets/regression_dataset.py and also create a regression metric class similar to https://github.com/Trusted-AI/AIF360/blob/master/aif360/metrics/classification_metric.py

@hoffmansc any thoughts?

Issue #263, opened 2021-08-23
Docker setup for AIF360


asm582:
Users may want to try AIF360 inside a docker container to launch on Kubernetes/Openshift environments.

asm582:
Hello, can you please help me get this PR merged?

nrkarthikeyan:
Hi @asm582 - what is in the zip file you have uploaded in the dataset?

asm582:
Hi @nrkarthikeyan the zip file is processed MEP dataset using Rscript ran on python2.7

nrkarthikeyan:
Ok. Can we provide instructions to create the zip file rather than actual
data? we do not want to include actual data.

PR #264, opened 2021-08-23
dockerize with dataset

asm582:
This PR solves #263. It provides users the ability to run AIF360 inside a docker container by adding datasets inside the container. This will be helpful to users launching containers on kubernetes/Openshift. 

Datasets are added in zip format and need to be unzipped in order to build the image. Such instructions are added in README.md file.


asm582:
Hello, Can you please help in getting this PR merged?

Dockerfile
```diff
@@ -0,0 +1,31 @@
+#base images
+FROM ubuntu:18.04
+FROM python:3.8
+
+#copy env file
+WORKDIR /src
+
+#install dependencies    
+RUN pip install aif360 && \
```
atinsood, line 9: you are install aif360 multiple times in this line.

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+#copy dataset
+COPY aif360/data/raw/meps/h181.csv /src/AIF360/aif360/data/raw/meps/h181.csv
+COPY aif360/data/raw/meps/h192.csv /src/AIF360/aif360/data/raw/meps/h192.csv
+
+
+#change work directory
+WORKDIR /src/AIF360
+
+#compile notebooks to run in jupyter lab
+RUN pip install --editable '.[all]' && \
```
atinsood, line 26: you are not debugging any code right, so you don't need the -e or --editable flag

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+    pip install -e '.[notebooks]' && \
+    pip install jupyterlab
```
atinsood, line 28: you can technically use a base container from jupyter and not ubuntu container and that will take care of you not needing to install jupyter yourself https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html

atinsood (commented):
minor comments, feel free to ignore and I can approve

nrkarthikeyan:
@asm582 - the dataset seems to be MEPS data which we do not want to have in the repo. Lets point them to https://github.com/Trusted-AI/AIF360/tree/master/aif360/data/raw/meps that shows how to create the csv files and provide instructions to create the zip file, rather than us providing the data.

asm582:
> @asm582 - the dataset seems to be MEPS data which we do not want to have in the repo. Lets point them to https://github.com/Trusted-AI/AIF360/tree/master/aif360/data/raw/meps that shows how to create the csv files and provide instructions to create the zip file, rather than us providing the data.

@nrkarthikeyan The idea here was to enable users to deploy AIF on Openshift/Kubernetes, without a sample dataset I am not sure if we have a working tutorial for users, do we have another dataset that we can distribute in containers?

Dockerfile
asm582: @atinsood readme instructions suggest using --editable for compilation, if the compilation fails I believe it should be verbose enough to help users figure out the issue.

Dockerfile
asm582: Thanks I have remove the duplicate install command from dockerfile

Issue #150, opened 2020-02-23
Port micellaneous items to sklearn-compatible API


hoffmansc:
- [ ] MEPS dataset
- [ ] differential fairness metrics

SSaishruthi:
@hoffmansc Do you skeleton for this conversion?

hoffmansc:
I don't really have skeletons for these but I thought examples would be enough:

For the MEPS dataset, a good place to start might be the [COMPAS port](https://github.com/IBM/AIF360/blob/master/aif360/sklearn/datasets/compas_dataset.py).

For differential fairness, you could look at any number of the [already ported metrics](https://github.com/IBM/AIF360/blob/master/aif360/sklearn/metrics/metrics.py).

SSaishruthi:
@hoffmansc I will take a look at it this week.

mkrueger12:
New to contributing. Can I take a crack at this?

krvarshney:
we'd love it if you did

mkrueger12:
Great. I'll give it a go.

hoffmansc:
Sorry! I got this confused with another issue. I have some work in progress for the first two but rich subgroup fairness is still outstanding. @mkrueger12

mkrueger12:
Sounds good. I can work on rich subgroup fairness. @hoffmansc

PR #278, opened 2021-11-23
Dataset Improvements

hoffmansc:
Major improvements:
- Changed how `prot_attr` arguments are handled. Now, when processing a dataset and running metrics, an explicit array (or list of arrays) containing protected attribute values per sample may be passed instead of requiring an index name.
- Added MEPS, COMPAS violent datasets (#150)

Small changes:
- Added `cache=False` option to dataset fetching functions to skip caching
- Removed unused categories from DataFrame resulting from dropped rows


nrkarthikeyan:
@monindersingh - FYI, this may address some of the questions you had.

hoffmansc:
@monindersingh I have a question about MEPS. In the pre-processing, you included `'EDUCYR'` and `'HIDEG'`:
https://github.com/Trusted-AI/AIF360/blob/963df2e4eea807bd5765fee9f1c594500bdcbb5b/aif360/datasets/meps_dataset_panel19_fy2015.py#L45
but in the `features_to_keep` arg of `__init__` they're not there:
https://github.com/Trusted-AI/AIF360/blob/963df2e4eea807bd5765fee9f1c594500bdcbb5b/aif360/datasets/meps_dataset_panel19_fy2015.py#L80-L85
Essentially this means those columns are included when we drop NAs but not in the final dataset. This only affects a few rows but what's the reasoning for this? Was it intentional?

PR #288, opened 2022-01-19
Fixed computation of coefficient of variation in classification_metrics

plankington:
The current implementation uses 2*sqrt(GE(alpha=2)); however, GE(2) = 1/2 (cv)^2 which implies sqrt(2*GE(2)) = cv.


krvarshney:
Good catch. Thanks for noticing it. @hoffmansc, are there any other downstream changes that would be needed beyond what @plankington has made?

hoffmansc:
> Good catch. Thanks for noticing it. @hoffmansc, are there any other downstream changes that would be needed beyond what @plankington has made?

Yep, updated now.

hoffmansc (approved):

PR #291, opened 2022-02-23
correct documented parameter which is not exposed

Deathn0t:
Hello,

I propose a tiny edit to the documentation, where I noticed that the `priv_group` parameter is documented in `sklearn.metrics.average_odds_error` but not accessible.

Best.


hoffmansc:
fixed by #318

PR #310, opened 2022-06-16
Added a tutorial for advertising data

barvek:


nrkarthikeyan (changes requested):
1. Add the dataset location from DAX and ask users to download and keep it in the same location as the notebook.
2. Under cell 19, modify the margin to 0.0055 in three places in the markdown

nrkarthikeyan (approved):

PR #313, opened 2022-06-20
fix rpy2 crash bug

hoffmansc:
rpy2 v3.5.2 appears to cause issues


hoffmansc:
appears to work now with `R` v4.2.0 and `rpy2` v3.5.2

PR #317, opened 2022-06-28
Remove caches due to excessive memory use

Adebayo-Oshingbesan:
The caches in the scoring functions lead to excessive memory usages for large datasets which sometimes led to the code crashing. Since the caches were intended to speedup runtime for those large datasets, they are not serving the intended purpose and thus needed to be removed.


hoffmansc (approved):
LGTM

Issue #154, opened 2020-02-24
Port pre-processing algorithms to sklearn-compatible API


hoffmansc:
- [ ] DisparateImpactRemover
- [ ] LearnedFairRepresentation
- [ ] OptimizedPreprocessing
- [X] Reweighing

jimbudarz:
Will this make it easier/unnecessary to convert back and forth between Pandas DataFrames and Binary Label Datasets, for example? I've been having issues with Reweighing, for example, as AIF360 tends to only work with numerical data but does not provide instructions for dummification while storing the metadata mappings to de-dummify later on.

hoffmansc:
Yes, this will allow DataFrames to be used directly with the algorithms. Reweighing is already implemented so you can try it out if you're comfortable using the development branch (`git checkout sklearn-compat`). It should be released in the latest stable version soon as well.

Do you mind explaining exactly what issues you were facing? Was it with `convert_to_dataframe()`?

jimbudarz:
Hi @hoffmansc, yes, I've been having trouble understanding how to use convert_to_dataframe() after creating my own BinaryLabelDataset. Perhaps it's my own fault, but I can't find the documentation that describes how to dummify the data in a way that retains the mappings so it can be reversed after using a PreProcessing tool such as Reweighing.

hoffmansc:
@jimbudarz, if you encode your categorical data with `pd.get_dummies()`, or use `StandardDataset`, you will end up with `feature_names` that look like, e.g., `[..., native-country=United-States, native-country=Vietnam, native-country=Yugoslavia, ...]`. Then, if you do `convert_to_dataframe(de_dummy_code=True)`, you will get a DataFrame that looks something like:
```
  ... native-country
0 ... United-States
1 ... United-States
2 ... Vietnam
...
```
with the columns magically mapped from one-hot to categories.

You can also include maps for the labels and protected attributes manually (since these are encoded differently) by supplying them when creating the BinaryLabelDataset (note: `protected_attribute_maps` should be in the same order as `protected_attribute_names`):
```
metadata = {
    'label_maps': [{1.0: '>50K', 0.0: '<=50K'}],
    'protected_attribute_maps': [{1.0: 'White', 0.0: 'Non-white'},
                                 {1.0: 'Male', 0.0: 'Female'}]
}
BinaryLabelDataset(..., metadata=metadata)
```
otherwise they will just be 0/1 which is probably also fine.

jimbudarz:
[...]

Issue #151, opened 2020-02-23
Port post-processing algorithms to sklearn-compatible API


hoffmansc:
- [X] CalibratedEqualizedOdds
- [ ] EqualizedOdds
- [ ] RejectOptionClassifier

PR #318, opened 2022-06-28
More sklearn-compatible algorithms

hoffmansc:
* RejectOptionClassification (#151)
* LearnedFairRepresentations (#154)
* option to suppress zero division warnings in metrics
* consolidate dataset loading in conftest.py


mnagired (approved):
both notebooks (LFR and ROC) look good to go :)

PR #338, opened 2022-08-25
Optional tempeh, conditional imports

DanielRyszkaIBM:
The following changes are introduced:
- making tempeh optional
- change imports from sklearn `deprecation` class
- conditional imports for libraries that are installed from extras


setup.py
```diff
@@ -22,6 +22,7 @@
     'Reductions': ['fairlearn==0.4.6'],
     'FairAdapt':['rpy2'],
     'notebooks': ['jupyter', 'tqdm', 'igraph[plotting]', 'lightgbm', 'seaborn', 'ipympl'],
+    'tempeh-datasets': ['tempeh'],
```
hoffmansc, line 25: can we call this `LawSchoolGPA` instead of `tempeh-datasets`?

aif360/sklearn/datasets/__init__.py
```diff
@@ -12,4 +12,7 @@
 from aif360.sklearn.datasets.openml_datasets import fetch_adult, fetch_german, fetch_bank
 from aif360.sklearn.datasets.compas_dataset import fetch_compas
 from aif360.sklearn.datasets.meps_datasets import fetch_meps
-from aif360.sklearn.datasets.tempeh_datasets import fetch_lawschool_gpa
+
+import pkgutil
+if pkgutil.find_loader("tempeh"):
+    from aif360.sklearn.datasets.tempeh_datasets import fetch_lawschool_gpa
```
hoffmansc, line 18: what is the benefit of this method instead of the try, except when actually importing tempeh?

hoffmansc (commented):
Thanks, this looks very good overall! Eventually we should remove the tempeh dependency altogether and directly download the dataset from SEAPHE but this is totally fine for now.

I left a couple comments in the code. Also, can you add the import warnings for the `aif360.sklearn.preprocessing` algos (FairAdapt and LFR)?

setup.py
DanielRyszkaIBM: Fixed in latest

aif360/sklearn/datasets/__init__.py
DanielRyszkaIBM: Yeah... no benefits over imports in try/except - changed in latest

hoffmansc (approved):

PR #404, opened 2022-09-16
Define "data stores" (to make it easy to download data from various sources)

anupamamurthi:
Signed-off-by: anupama.murthi@gmail.com

### DIFF
--- a/aif360/stores/kaggle_store.py
+++ b/aif360/stores/kaggle_store.py
@@ -1,0 +1,36 @@
+import abc
+import os
+
+from aif360.stores.store import Store
+from aif360.utils import get_logger
+
+# To use Kaggle store,
+# 1. Login to Kaggle
+# 2. Create new API token
+# 3. Save Kaggle.json to ~/.kaggle
+# 4. Use KaggleStore
+
+
+class KaggleStore(Store):
+
+    def __init__(self, source, destination):
+
+        self.logger = get_logger(type(self).__name__)
+        self.validate_store(source)
+        self.source = source
+        self.destination = destination
+
+    def validate_store(self, dataset):
+        # TODO: check if dataset actually exists in kaggle?
+        pass
+
+    def download(self):
+        # importing kaggle inside download() to avoid an authentication error.
+        import kaggle
+        kaggle.api.authenticate()
+        kaggle.api.dataset_download_files(self.source, path=self.destination,
+                                          unzip=True)
+        return len(os.listdir(self.destination)) != 0
+
+    def existsInDestination(self, **kwargs):
+        pass
--- a/aif360/stores/store.py
+++ b/aif360/stores/store.py
@@ -1,0 +1,27 @@
+import abc
+import copy
+import sys
+import abc
+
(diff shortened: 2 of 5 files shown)


aif360/stores/kaggle_store.py
```diff
@@ -0,0 +1,36 @@
...
+from aif360.utils import get_logger
+
+# To use Kaggle store,
+# 1. Login to Kaggle
+# 2. Create new API token
+# 3. Save Kaggle.json to ~/.kaggle
+# 4. Use KaggleStore
+
+
+class KaggleStore(Store):
```
anupamamurthi, line 14: Will add a little usage block + sample in a follow up PR

aif360/stores/kaggle_store.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def __init__(self, source, destination):
+
+        self.logger = get_logger(type(self).__name__)
+        self.validate_store(source)
+        self.source = source
+        self.destination = destination
+
+    def validate_store(self, dataset):
```
aroma123, line 23: Perhaps, might want to verify md5 checksum of the downloaded data.

hoffmansc:
A couple thoughts:

- Can these just be functions? Is there a use case that requires it to be a class? This would make it more in line with `fetch_openml`
- Pandas can download from a URL natively (e.g. [read_csv](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html?highlight=URL)). The drawback is it won't cache the file. It also doesn't work with zipped files...
- A lot of the utils seem unused. Are they necessary?

anupamamurthi:
Hi @hoffmansc this is my take..
1) Classes maybe handy to do things like -validate to see if the incoming UCI URL or Kaggle URL is valid and we can extend these classes in the future (if needed) by adding additional error checks that are specific to the "data store".
For COS, I can think of how we may want to "list cos objects" or get a particular version of the cos object instead of getting the latest etc. Maybe too futuristic.
but one could argue saying we can live with functions (similar to fetch_openml). I am fine with that as well. 

regarding 3rd point, yeah I thought these utils may come handy but I can remove them and can be added as and when needed!

hoffmansc:
I guess functions seems simpler to me. I'm worried this is a little overcomplicated. For example, it seems like the kaggle function essentially does what we need (assuming it checks if the dataset is valid or already exists). Versioning and error checks can still happen inside the function. Listing available datasets could be a separate function. The only thing is if authentication gets cumbersome to input but, at least for kaggle, it's stored in a file so this isn't an issue.

anupamamurthi:
I hear you and Karthi shared the same concern :) I will fix it up. Will have some util functions instead!

hoffmansc:
Another thing I thought of is we only check if the file names are present in the cache dir but it's possible two URLs have the same file names but different contents. In other words, the cache "key" should be the URL+file. Is this an issue or should we ignore this unlikely problem? Maybe we could hash the URL somehow and save the files in a folder under that unique name? It sacrifices readability though

Issue #292, opened 2022-03-04
type(...) instead of isinstance(...)


alfa-yohannis:
https://github.com/Trusted-AI/AIF360/blob/48acf62c827a00dff881d4b08903fee2d108a9b9/aif360/explainers/metric_json_explainer.py#L155

I think there is a logic error in `disparate_impact()` and  `statistical_parity_difference()`. The line `if isinstance(self.metric, BinaryLabelDatasetMetric):` should be `if type(self.metric) is BinaryLabelDatasetMetric:`, otherwise any class derived from `BinaryLabelDatasetMetric`, including `ClassificationMetric`, always goes to the first condition.

Cheers,

PR #407, opened 2022-09-16
 logic error in disparate_impact and statistical_parity_difference #292

sreeja-g:
#292 fix

type() instead of isinstance()


hoffmansc (approved):
LGTM, thanks!

nrkarthikeyan (approved):
Can commit after tests pass.

Issue #376, opened 2022-09-14
implement average predictive value difference metric


nrkarthikeyan:
See eqn 10.4 in http://trustworthymachinelearning.com/trustworthymachinelearning-10.htm

First priority is to do this for sklearn compatible version and then for "classic" version.

PR #410, opened 2022-09-16
average predictive value difference metric implementation #376

sreeja-g:
#376 implementation

average predictive value difference metric

### DIFF
--- a/aif360/sklearn/metrics/metrics.py
+++ b/aif360/sklearn/metrics/metrics.py
@@ -344,7 +344,7 @@ def specificity_score(y_true, y_pred, *, pos_label=1, sample_weight=None,
     return _prf_divide(tn, negs, 'specificity', 'negative', None,
                        ('specificity',), zero_division).item()
 
-def false_omission_score(y_true, y_pred, *, pos_label=1, sample_weight=None,
+def false_omission_rate_error(y_true, y_pred, *, pos_label=1, sample_weight=None,
                       zero_division='warn'):
     """Compute the false omission rate.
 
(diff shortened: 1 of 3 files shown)


aif360/sklearn/metrics/metrics.py
```diff
@@ -344,6 +344,27 @@ def specificity_score(y_true, y_pred, *, pos_label=1, sample_weight=None,
     return _prf_divide(tn, negs, 'specificity', 'negative', None,
                        ('specificity',), zero_division).item()
 
+def false_omission_score(y_true, y_pred, *, pos_label=1, sample_weight=None,
```
hoffmansc, line 347: By [scikit-learn convention](https://scikit-learn.org/stable/modules/model_evaluation.html#defining-your-scoring-strategy-from-metric-functions), functions ending in `_score` should be maximized while functions ending in `_error` should be minimized. Can we change this name to `false_omission_rate_error`?

aif360/sklearn/metrics/metrics.py
```diff
@@ -344,6 +344,27 @@ def specificity_score(y_true, y_pred, *, pos_label=1, sample_weight=None,
...
+            zero division. If set to “warn”, this acts as 0, but warnings are
+            also raised.
+    """
+    _check_zero_division(zero_division)
+    MCM = multilabel_confusion_matrix(y_true, y_pred, labels=[pos_label],
+                                      sample_weight=sample_weight)
+    tn, fn = MCM[:, 0, 0], MCM[:, 1, 0]
+    negs = tn + fn
+    return _prf_divide(fn, negs, 'negative_predictive', 'negative', None,
+                       ('negative_predictive',), zero_division).item()
```
hoffmansc, line 366: The error message should read: ```UndefinedMetricWarning: False Omission Rate is ill-defined and being set to 0.0 due to no predicted negative samples. Use `zero_division` parameter to control this behavior.```

Therefore, this line should be:
```
    return _prf_divide(fn, negs, 'false omission rate', 'predicted negative',
                       None, ('false omission rate',), zero_division).item()
```

hoffmansc (changes requested):
Great work! Just a couple small comments and we should be good to go!

hoffmansc (approved):
Looks great, thanks!

PR #412, opened 2022-09-17
Add code coverage checks with `pytest-cov`

### LINKED ISSUE #397: Enable code coverage with Pytest with PyTest-Cov
Enable Coverage computation using PyTest-Cov in CI. See  https://pypi.org/project/pytest-cov/

aitorres:
closes #397

## What I did

- [x] Added `pytest-cov` to the project dependencies
- [x] Enabled coverage checks when running `pytest` and set a minimum of 80% coverage (this can be changed if needed)

Let me know if there's anything else I could add or fix to this PR and I'll gladly help!

## How to test

We should be able to check if this works by looking at this PR's Github Actions run!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
           flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
 
       - name: Test with pytest
-        run: pytest tests
+        run: pytest tests --cov=aif360 --cov-report=term-missing --cov-fail-under=80
 
   build-r:
     runs-on: ubuntu-18.04
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
           flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
 
       - name: Test with pytest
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -67,7 +67,7 @@ jobs:
           flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
 
       - name: Test with pytest
-        run: pytest tests
+        run: pytest tests --cov=aif360 --cov-report=term-missing --cov-fail-under=80
```
hoffmansc, line 70: Let's start small and just report the coverage instead of setting a failure criterion -- mainly because we're failing right now 🙂

Eventually we can add it back and/or upload to codecov once we add more tests

hoffmansc (changes requested):
Great work, thanks!

It looks like you also need to add pytest-cov to [setup.py](https://github.com/Trusted-AI/AIF360/blob/master/setup.py#L27), though

.github/workflows/ci.yml
aitorres: Thanks for the comments! I just pushed a couple commits adding `pytest-cov` to `setup.py` and removing the failure criterion.

hoffmansc (approved):
Thanks for the fixes! Looks great

PR #415, opened 2022-10-07
Remove turtle module import

gkumbhat:
### Changes
- Remove `turtle` module import, since it doesn't seem to be getting used and breaks aif360

Signed-off-by: Gaurav Kumbhat <kumbhat.gaurav@gmail.com>


hoffmansc (approved):

lanterno:
Thank you! 
indeed, this import was breaking the library for us. causing some weird import to tkinter which was not available on our ML python kernels.

lanterno:
@hoffmansc Can we get a new patch release with this PR please? 🙏 

I tried installing the library from source into my own project, but faced lots of different issues for that.

katkamrachana:
@hoffmansc following up on releasing new version with the fix. TIA!

PR #434, opened 2023-01-22
Add a bias detector based on optimal transport

### LINKED ISSUE #433: Add a bias detector based on optimal transport
There is an increasing interest in estimating bias in terms of Wasserstein-2 distances (or related distances between measures, often computable using optimal-transport algorithms), cf.
http://proceedings.mlr.press/v97/gordaliza19a.html (ICML 2019)
https://academic.oup.com/imaiai/article-abstract/8/4/817/5586771 (Information and Inference, 2019)
https://openreview.net/forum?id=-welFirjMss (NeurIPS 2022)
Following some discussions with Rahul Nair (rahul.nair@ie.ibm.com), we would like to contribute a new detector, with the same signature as the original mdss detector of @Adebayo-Oshingbesan:
https://github.com/Trusted-AI/AIF360/blob/master/aif360/detectors/mdss_detector.py
with Illia Kryvoviaz.

A draft implementation is at:
https://github.com/Trusted-AI/AIF360/compare/master...Illia-Kryvoviaz:AIF360:master

jmarecek:
There is an increasing interest in estimating bias in terms of Wasserstein-2 distances (or related distances between measures, often computable using optimal-transport algorithms), cf.
http://proceedings.mlr.press/v97/gordaliza19a.html (ICML 2019)
https://academic.oup.com/imaiai/article-abstract/8/4/817/5586771 (Information and Inference, 2019)
https://openreview.net/forum?id=-welFirjMss (NeurIPS 2022)
Following some discussions with Rahul Nair ([rahul.nair@ie.ibm.com](mailto:rahul.nair@ie.ibm.com)), we would like to contribute a new detector, with the same signature as the original mdss detector of @Adebayo-Oshingbesan:
https://github.com/Trusted-AI/AIF360/blob/master/aif360/detectors/mdss_detector.py

Could we ask @hoffmansc for a code review, please? 

Closes #433


jmarecek:
Hi @hoffmansc, 

I hope all is well. Illia has pushed the unit tests, doc strings, etc. It would be great to revisit this. Illia and I would be happy to jump on a call, should this be faster?

Thanks!
Jakub

rahulnair23:
Hi @jmarecek @Illia-Kryvoviaz - I've reviewed this as well and happy to have a call to discuss some details here. In addition to the points raised above by @hoffmansc (FYI - point 1 is still open, and points 2 and 3 have only been partially completed), please also consider: 

- [ ] The output of the bias detector [here](https://github.com/Illia-Kryvoviaz/AIF360/blob/727196405f024694a74d1134a50f2d36803ecc92/aif360/detectors/ot_detector.py#L229) appears to only return the cost of the transport. This in itself lacks context. Could you include additional context, for example, similar to the output of MDSS scan?
- [ ] The demo notebook [here](https://github.com/Illia-Kryvoviaz/AIF360/blob/master/examples/demo_ot_detector.ipynb) contains the references to the MDSS paper. Is there a reference paper for use of optimal transport for bias detection you can cite here? 
- [ ] The demo notebook on the COMPASS dataset ends with the cost of the transport which isn't illustrative as a bias scan. Are the additional steps you can include to show use?

Thank you.

Illia-Kryvoviaz:
Hi @hoffmansc!
We have taken into account your and @rahulnair23's suggestions, and now they have been implemented.
Could we ask you for a code review, please?

https://github.com/Illia-Kryvoviaz/AIF360/blob/master/aif360/detectors/ot_detector.py

- [ ] Can you include an ot_bias_scan function in aif360/sklearn/detectors/ as well? The functionality should be the same but the signature is a bit different. This is just for consistency. See [aif360/sklearn/detectors/detectors.py](https://github.com/Trusted-AI/AIF360/blob/master/aif360/sklearn/detectors/detectors.py) for an example.
--> https://github.com/Illia-Kryvoviaz/AIF360/blob/master/aif360/sklearn/detectors/detectors.py

- [ ] Can you import your function in the __init__.py files for aif360/detectors/ and aif360/sklearn/detectors/? Again, for consistency. This will allow us to import the function directly from the subpackage instead of the source file.
--> https://github.com/Illia-Kryvoviaz/AIF360/blob/master/aif360/detectors/__init__.py
--> https://github.com/Illia-Kryvoviaz/AIF360/blob/master/aif360/sklearn/detectors/__init__.py

- [ ] I think you'll need to add ot to the dependency list. You can do this by adding an entry to the extras dict in [setup.py](https://github.com/Trusted-AI/AIF360/blob/master/setup.py) (you can name it 'OptimalTransport') as well as [requirements.txt](https://github.com/Trusted-AI/AIF360/blob/master/requirements.txt).
--> https://github.com/Illia-Kryvoviaz/AIF360/blob/master/setup.py
--> https://github.com/Illia-Kryvoviaz/AIF360/blob/master/requirements.txt

- [ ] We also need a unit test file for this PR. Please see the [tests section of the contribution guide](https://github.com/Trusted-AI/AIF360/blob/c9a542b862cf988cbc788871c6cc7f2a926dad31/CONTRIBUTING.rst#tests) for more information.
--> https://github.com/Illia-Kryvoviaz/AIF360/blob/master/tests/test_ot_detector.py

- [ ] For the demo notebook, could you add some text explaining what the ot_matrix means, how it should be used, etc.? Also, can you make it clear what the difference between this and MDSS bias scan is in the text? The description just seems copy-pasted from the other notebook.
--> https://github.com/Illia-Kryvoviaz/AIF360/blob/master/examples/demo_ot_detector.ipynb

aif360/sklearn/detectors/detectors.py
```diff
@@ -1,10 +1,76 @@
...
+    y_pred: Union[pd.Series, pd.DataFrame, str],
+    sensitive_attribute: Union[pd.Series, str] = None,
+    X: pd.DataFrame = None,
+    pos_label: Union[str, float] = None,
+    overpredicted: bool = True,
+    scoring: str = "Optimal Transport",
+    num_iters: int = 100,
+    penalty: float = 1e-17,
+    mode: str = "ordinal",
+    **kwargs,
```
hoffmansc, line 20: same as other comment -- no need to include unused args -- however, in this case, it's good to have the same ordering as `bias_scan`:
`X, y_true, y_pred, prot_attr, pos_label, ...`

aif360/sklearn/detectors/detectors.py
```diff
@@ -1,10 +1,76 @@
...
 from aif360.detectors.mdss.ScoringFunctions import ScoringFunction
 
+from typing import Union
 import pandas as pd
+import numpy as np
+
+def ot_bias_scan(
+    y_true: Union[pd.Series, str],
+    y_pred: Union[pd.Series, pd.DataFrame, str],
+    sensitive_attribute: Union[pd.Series, str] = None,
```
hoffmansc, line 12: can we rename this to `prot_attr`?

examples/demo_ot_detector.ipynb
```diff
@@ -0,0 +1,1403 @@
...
+       "4               1-39                0                0         0     0.236226  "
+      ]
+     },
+     "execution_count": 22,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "data = pd.read_csv('https://gist.githubusercontent.com/Viktour19/b690679802c431646d36f7e2dd117b9e/raw/d8f17bf25664bd2d9fa010750b9e451c4155dd61/adult_autostrat.csv')\n",
```
hoffmansc, line 1078: why do you use this version of adult instead of the normal one?

examples/demo_ot_detector.ipynb
```diff
@@ -0,0 +1,1403 @@
...
+      "\n",
+      "Total bias = 0.00026458993506953317\n"
+     ]
+    }
+   ],
+   "source": [
+    "bs2 = pd.DataFrame({\"age_cat\": ot_val2.keys(), \"ot_val\": ot_val2.values()})\n",
+    "display(bs2)\n",
+    "\n",
+    "print(\"\\nTotal bias = \", abs(np.diff(list(ot_val2.values()))[0]), sep=\"\")"
```
hoffmansc, line 881: this is just computing the difference between `age_cat` 0 and 1. Is that intentional(/meaningful)?

hoffmansc (commented):
Thanks @Illia-Kryvoviaz and @jmarecek for all the hard work, this is looking really good! The notebook in particular is very informative.

I have one big picture question, though, now that I understand what's happening better. I'm a bit confused why this is classified as a detector and not just a metric. MDSS bias scan is sort of special in that it identifies biased subgroups not determined a priori by sensitive attributes but this seems to be more similar to methods in the [metrics](https://github.com/Trusted-AI/AIF360/blob/master/aif360/sklearn/metrics/metrics.py) subpackage. This seems to be more of a typical group fairness metric from what I can tell. Let me know if I'm misunderstanding.

Also, just for clarity, is this implementing one specific paper from the references?

aif360/detectors/ot_detector.py
```diff
@@ -0,0 +1,257 @@
...
+    classifier: Union[pd.Series, str],
+    sensitive_attribute: Union[pd.Series, str] = None,
+    data: pd.DataFrame = None,
+    favorable_value: Union[str, float] = None,
+    overpredicted: bool = True,
+    scoring: str = "Optimal Transport",
+    num_iters: int = 1e5,
+    penalty: float = 1e-17,
+    mode: str = "binary",
+    **kwargs,
```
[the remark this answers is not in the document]
Illia-Kryvoviaz: `overpredicted` is indeed unnecessary - will remove.
We do have plans for future `penalty` and `scoring` options (The multiplier associated with the ergodic term in the Sinkhorn algorithm, and Wasserstein-1 and Wasserstein-2 options, respectively)
We suggest to keep `kwargs` as a way to incorporate the less used parameters that may arise in the future. This is the case with `cost_matrix`, as it is rarely necessary, but can still be useful in some cases.

examples/demo_ot_detector.ipynb
Illia-Kryvoviaz: It was the one used in the MDSS example notebook - will change to use `load_preproc_data_adult`

examples/demo_ot_detector.ipynb
Illia-Kryvoviaz: Indeed, this doesn't offer much information in our case, will be removed

aif360/detectors/ot_detector.py
```diff
@@ -0,0 +1,257 @@
...
+def _evaluate(
+        ground_truth: pd.Series,
+        classifier: pd.Series,
+        sensitive_attribute: pd.Series=None,
+        data: pd.DataFrame=None,
+        num_iters=1e5,
+        **kwargs):
+    """If the given golden_standart and classifier are distributions, it returns the Wasserstein distance between them, 
+    otherwise it extract all neccessary information from data, makes logistic regression and 
+    compute optimal transport for all possible options for the given classifier.
```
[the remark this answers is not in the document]
Illia-Kryvoviaz: this comment needs updating, thank you!

Illia-Kryvoviaz:
Hello @hoffmansc!
We updated the code based on your feedback. Could you please check out the changes and the [argument ordering question](https://github.com/Trusted-AI/AIF360/pull/434#discussion_r1256478154)?

As for detector vs metric, we argue that this should be an independent detector, rather than a metric, because it is too time-consuming to be a metric or run the subgroup analysis similar to MDSS.

aif360/detectors/ot_detector.py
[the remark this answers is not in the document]
hoffmansc: I see so `cost_matrix` is specific to this `scoring` option? If other scoring options don't need this then it's fine to keep it under `kwargs` but if it's universal for any scoring function, just rare, it should be named. The only reason to have `kwargs` is to pass `scoring`-specific args which we can't know ahead of time since there are multiple possible scorers.

aif360/sklearn/detectors/detectors.py
hoffmansc: I see, so pass the features is optional. In that case it's fine as is although I don't really see a scenario where it's easier to pass `X` and strings for `y_true`, `y_pred`, etc. since you're not actually using the features in the algorithm. `prot_attr` is often a str because in `aif360/sklearn/` we store them in the index of the `pd.Series` or you could pass an array directly but I think maybe it's more confusing to combine features and labels and call it `X`.

Do you feel strongly about keeping the `X`/`data` option or can we just drop it and only pass arrays/Series?

hoffmansc:
Thanks for your updates. I'm not sure I understand your point about detector vs. metric. It's fine if the metric is slow. The question is which classification better fits the algorithm and I think if it doesn't identify a subgroup like MDSS it shouldn't be called a detector. Metrics are simply anything which takes in data (predictions, ground truth, features, groups, etc.) and returns a numerical score (either per-group or "reduced" by taking the difference between groups, for example). Am I misinterpreting your algorithm?

jmarecek:
Like I mentioned above, there are a variety of scoring functions within the "Optimal transport perspective", incl. 
https://en.wikipedia.org/wiki/Wasserstein_metric Wp for integers p,
https://pythonot.github.io/auto_examples/gromov/plot_gromov.html for various integers and metrics on the underlying spaces,
fused Gromov-Wasserstein variants, etc. So I don't really think it is a single scoring function.

aif360/sklearn/detectors/detectors.py
Illia-Kryvoviaz: dropped the `X` option in the new pull request

hoffmansc:
I understand there can be multiple distance/scoring functions. I'm not saying there needs to be any changes to the function just that it would be better suited under `aif360.metrics` instead of `aif360.detectors`. Even though it has options with respect to scoring function, the output is still a numerical value not a subgroup.

krvarshney:
Sam, your logic for it to be under metrics makes sense to me.

jmarecek:
Hi Kush -- so what about this: we contribute the OT-based metrics to aif360.metrics, and we add some "SimpleDetector" to detectors, which would not try to run the subgroup scan. That way, the current functionality of the code would be easy to access. Would that work? Jakub

rahulnair23:
Thanks for the call and discussion today @hoffmansc @jmarecek . As concluded, from an end-user point of view the OT-based measures are better suited under `aif360.metrics` (as opposed to `aif360.detectors`).

hoffmansc:
Actually, to be clear, it's only strictly necessary to put it under `aif360.sklearn.metrics`. This is the most straightforward thing to implement. It is optional to put it under `aif360.metrics` (i.e., the class-based Metric interface). The scikit-learn style functions are preferred going forward and the Metric class is only supported for backwards compatibility.

aif360/metrics/ot_metric.py
```diff
@@ -0,0 +1,225 @@
...
+    for sa_val in sorted(prot_attr.unique()):
+        initial_distribution = ground_truth[prot_attr == sa_val]
+        required_distribution = classifier[prot_attr == sa_val]
+        initial_distribution, required_distribution, matrix_distance = _transform(initial_distribution, required_distribution, cost_matrix)
+        emds[sa_val] = ot.emd2(a=initial_distribution, b=required_distribution, M=matrix_distance, numItermax=num_iters)
+
+    return emds
+
+
+def ot_bias_scan(
```
hoffmansc, line 93:
```suggestion
def ot_distance(
```

aif360/sklearn/detectors/detectors.py
```diff
@@ -1,10 +1,9 @@
-from typing import Union
-
 from aif360.detectors import bias_scan
 from aif360.detectors.mdss.ScoringFunctions import ScoringFunction
 
+from typing import Union
 import pandas as pd
-
+import numpy as np
```
hoffmansc, line 6: you can reset these changes -- there's no need to edit this file anymore

aif360/sklearn/metrics/metrics.py
```diff
@@ -24,7 +27,7 @@
     'specificity_score', 'base_rate', 'selection_rate', 'smoothed_base_rate',
     'smoothed_selection_rate', 'generalized_fpr', 'generalized_fnr',
     # group fairness
-    'statistical_parity_difference', 'disparate_impact_ratio',
+    'ot_bias_scan', 'statistical_parity_difference', 'disparate_impact_ratio',
```
hoffmansc, line 30:
```suggestion
    'ot_distance', 'statistical_parity_difference', 'disparate_impact_ratio',
```

aif360/sklearn/metrics/metrics.py
```diff
@@ -499,6 +502,67 @@ def generalized_fnr(y_true, probas_pred, *, pos_label=1, sample_weight=None,
...
+    """Normalize and calculate Wasserstein distance between groups defined by `prot_attr` in `y_true` and `y_pred`.
+
+    Args:
+        y_true (pd.Series): ground truth (correct) target values.
+        y_pred (pd.Series, pd.DataFrame): estimated target values.
+            If `mode` is nominal, must be a `pd.DataFrame` with columns containing predictions for each nominal class,
+                or list of corresponding column names in `data`.
+            If `None`, model is assumed to be a dummy model that predicts the mean of the targets
+                or 1/(number of categories) for nominal mode.
+        sensitive_attribute (pd.Series): sensitive attribute values.
```
hoffmansc, line 527:
```suggestion
        prot_attr (pd.Series): sensitive attribute values.
```

aif360/sklearn/metrics/metrics.py
```diff
@@ -499,6 +502,67 @@ def generalized_fnr(y_true, probas_pred, *, pos_label=1, sample_weight=None,
...
+            If `None`, assume all samples belong to the same protected group.
+        pos_label(str, float, optional): Either "high", "low" or a float value if the mode in [binary, ordinal, or continuous].
+                If float, value has to be the minimum or the maximum in the ground_truth column.
+                Defaults to high if None for these modes.
+                Support for float left in to keep the intuition clear in binary classification tasks.
+                If `mode` is nominal, favorable values should be one of the unique categories in the ground_truth.
+                Defaults to a one-vs-all scan if None for nominal mode.
+        overpredicted (bool, optional): flag for group to scan for.
+            `True` scans for overprediction, `False` scans for underprediction.
```
hoffmansc, line 536:
```suggestion
```

aif360/sklearn/metrics/metrics.py
```diff
@@ -499,6 +502,67 @@ def generalized_fnr(y_true, probas_pred, *, pos_label=1, sample_weight=None,
...
+        prot_attr=prot_attr,
+        favorable_value=pos_label,
+        overpredicted=overpredicted,
+        scoring=scoring,
+        num_iters=num_iters,
+        penalty=penalty,
+        mode=mode,
+        cost_matrix=cost_matrix,
+        kwargs=kwargs
+    )
```
hoffmansc, line 564:
```suggestion
    return ot_metric.ot_distance(
        ground_truth=y_true,
        classifier=y_pred,
        prot_attr=prot_attr,
        favorable_value=pos_label,
        scoring=scoring,
        num_iters=num_iters,
        penalty=penalty,
        mode=mode,
        cost_matrix=cost_matrix,
        **kwargs
    )
```

hoffmansc (changes requested):
This looks good for now, thanks. I made one suggestion -- renaming the function to `ot_distance`. If this is acceptable, I can merge it.

Down the line, I would like this to integrate with `aif360.sklearn.metrics.intersection()`/`difference()` instead of how it handles `prot_attr` currently but I can make a separate PR with those changes and ask for your review.

Illia-Kryvoviaz:
Hello @hoffmansc! 
All changes were implemented as you suggested.

hoffmansc:
@Illia-Kryvoviaz It looks like one of the tests is failing. This line should be `pos_label=fav`
https://github.com/Trusted-AI/AIF360/blob/bb2ec68cd833e41fa73d06e20686cef0bc3eedef/tests/test_ot_metric.py#L136

Illia-Kryvoviaz:
Yes, you are right @hoffmansc. Now it looks right.

hoffmansc (approved):

PR #447, opened 2023-02-13
added demv preprocessing algorithm

giordanoDaloisio:
We included an implementation of the Debiaser for Multiple Variables algorithm from the paper

[Debiaser for Multiple Variables to enhance fairness in classification tasks](https://doi.org/10.1016/j.ipm.2022.103226)
G d’Aloisio, A D’Angelo, A Di Marco, G Stilo
Information Processing & Management 60 (2), 103226

We included the implementation of the algorithm, a notebook showing an example usage and also an implementation of a multiclass dataset for future use (CMC dataset)

Signed-off-by: Giordano d'Aloisio [giordano.daloisio@graduate.univaq.it](mailto:giordano.daloisio@graduate.univaq.it)


hoffmansc:
Hi @giordanoDaloisio, thanks for your contribution! I have a couple quick suggestions:

- [ ] I think the notebook can be improved. Right now, it doesn't really explain clearly what's happening at each step. Users can always read the paper to understand exactly what's going on but the notebook should at least give a basic overview, explain the advantages compared to other algorithms, demonstrate interesting behavior, etc.
- [ ] There are no assert statements in the notebook so the test file isn't really doing much other than testing for runtime errors. In any case, I would suggest a true unit test file independent of the demo that more thoroughly tests edge cases, etc.
- [ ] Finally, it would be really appreciated if you could include a version compatible with the `aif360.sklearn` API as well. It looks like you're using DataFrames internally anyway so I hope this isn't too much trouble.

giordanoDaloisio:
Hi @hoffmansc, here is the implementation of DEMV for the sklearn api

Issue #451, opened 2023-03-31
Typo in grid_search_reduction.py causing AttributeError: 'GridSearchReduction' object has no attribute 'moment'


haas-christian:
In the current version of [GridSearchReduction,](https://github.com/Trusted-AI/AIF360/blob/master/aif360/algorithms/inprocessing/grid_search_reduction.py) there's a typo in the predict() function that causes an AttributeError when trying to use a fitted GridSearchReduction model for predicting a new dataset.

Specifically, the typo is in line 140:
`if isinstance(self.model.moment, red.ClassificationMoment):` 

Since the model does not have an attribute moment, it will cause the Attribute Error. Instead, the moment attribute of the model should be self.model.moment_ :
`if isinstance(self.model.moment_, red.ClassificationMoment):` 

Here's an example that shows the described behavior:

```
# Load all necessary packages
from aif360.algorithms.preprocessing.optim_preproc_helpers.data_preproc_functions import load_preproc_data_adult
from aif360.algorithms.inprocessing.grid_search_reduction import GridSearchReduction
from sklearn.linear_model import LogisticRegression
import numpy as np

# Get the dataset and split into train and test
dataset_orig = load_preproc_data_adult()

privileged_groups = [{'sex': 1}]
unprivileged_groups = [{'sex': 0}]

np.random.seed(0)
dataset_orig_train, dataset_orig_test = dataset_orig.split([0.7], shuffle=True)

estimator = LogisticRegression(solver='lbfgs', max_iter=1000)

np.random.seed(0) #need for reproducibility
grid_red = GridSearchReduction(estimator=estimator,
                               constraints="EqualizedOdds",
                               drop_prot_attr=False)
grid_red.fit(dataset_orig_train)
grid_red_pred = grid_red.predict(dataset_orig_test)
```

haas-christian:
That's a good question. Based on my understanding, we want to check if the provided moment is one that is implemented as ClassificationMoment in fairlearn. `if isinstance(self.model.moment_, red.ClassificationMoment)` would check this, since self.moment_ in the sklearn-compatible version of GridSearchReduction refers to the fairlearn ClassificationMoment. In contrast, if we call `self.model.constraints`, we only get the str representation. 

In the above code sample, when I call self.model.moment_, I get the result `<fairlearn.reductions._moments.utility_parity.EqualizedOdds at 0x1bb41350130>`, indicating it's a FairLearn moment. 
[...]

PR #452, opened 2023-04-05
Fix typo in GridSearchReduction

haas-christian:
Linked to issue #451 

The previous version of predict() in [GridSearchReduction](https://github.com/Trusted-AI/AIF360/blob/master/aif360/algorithms/inprocessing/grid_search_reduction.py) had a typo where it tried to call `self.model.moment` to check which ClassificationMoment the model uses. 

This should be `self.model.moment_` in accordance with the sklearn-compatible version of [GridSearchReduction](https://github.com/Trusted-AI/AIF360/blob/master/aif360/sklearn/inprocessing/grid_search_reduction.py).

### DIFF
--- a/aif360/algorithms/inprocessing/grid_search_reduction.py
+++ b/aif360/algorithms/inprocessing/grid_search_reduction.py
@@ -137,7 +137,7 @@ class GridSearchReduction(Transformer):
         dataset_new = dataset.copy()
         dataset_new.labels = self.model.predict(X_df).reshape(-1, 1)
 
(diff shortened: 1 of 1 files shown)


mnagired:
@hoffmansc, this should be a quick fix which resolves a small issue with GridSearchReduction the base AIF360 library, documented in #451

hoffmansc (approved):
We should add a test for this as well, ideally, but that can be separate. Apparently we only run `ExponentiatedGradientReduction`.

richeekAbacus:
Can this be merged?

PR #457, opened 2023-06-29
Fairness Aware Counterfactuals for Subgroups

### LINKED ISSUE #456: Add implementation for Fairness Aware Counterfactuals for Subgroups
We have implemented the "FACTS" framework from [our respective paper](https://arxiv.org/abs/2306.14978). We believe that it would make a useful addition to the fairness auditing methods already provided by aif360.

(Placeholder issue for the respective pull request)

phantom-duck:
@hoffmansc 

Hello! We would like to contribute the implementation for the paper ["Fairness Aware Counterfactuals for Subgroups"](https://arxiv.org/abs/2306.14978).

Fairness Aware Counterfactuals for Subgroups (FACTS), a framework for auditing subgroup fairness through counterfactual explanations.  We aim to (a) formulate different aspects of the difficulty of individuals in certain subgroups to achieve recourse, i.e. receive the desired outcome, either at the micro level, considering members of the subgroup individually, or at the macro level, considering the subgroup as a whole, and (b) introduce notions of subgroup fairness that are robust, if not totally oblivious, to
the cost of achieving recourse.

For the implementation, we expose to the user a set of functions which perform the main steps of our algorithm and which we have grouped in the wrapper class FairnessAwareSubgroupCounterfactuals in the `facts` submodule's `__init__`.

Closes #456


phantom-duck:
Amended last commit - had forgotten to sign off.

We are looking forward to your feedback!

phantom-duck:
@hoffmansc 

Hello again! I would just like to remind you, when you get a chance, to take a look at the pull request and inform us of any corrections or improvements that may be needed; or merge the pull request if you believe it is adequate. In any case, we are looking forward to your feedback!

hoffmansc:
Hi. First of all, thanks for contributing, this is great stuff! I have a couple comments before we get into the specifics of the code.

I see you’ve placed the code under `algorithms.postprocessing`. This subpackage is for bias mitigation algorithms which transform predictions and _return fairer predictions_. To me, this seems different than what FACTS is doing. I think there should be three places where the contributions from this work get exposed: `sklearn.detectors`, `sklearn.metrics`, and `sklearn.explainers` (as a side note, we can focus on the `aif360.sklearn` subpackage, which will be much more straightforward for this work).

`explainers` is the clearest match in my mind. Currently, there is one other PR under review that will also go here (#450). Basically, your whole reporting function can go here. The idea here is to offer textual or visual explanations for bias measurements/metrics.

Correspondingly, I think there should be a function under `metrics` which just outputs the unfairness score as a number, i.e., contribution (b) you mention. This could either be the highest score or possibly given a specific subgroup. This way, we can easily compare classifiers directly.

This brings me to the final location, `detectors`. We currently have one detector, MDSS, and the idea behind this class is to discover subgroups which are maximally advantaged/disadvantaged. From my understanding of FACTS, you similarly generate subgroups and further divide them according to a protected attribute. The outputs here would simply be the subgroup and score, not the entire explanation.

Essentially, I’m trying to think about the various ways a user might want to utilize this work as well as how they would discover it according to the taxonomy we have in AIF360.

Feel free to disagree if I’m misunderstanding anything. Once we settle these high-level questions, I’ll take a closer look at the code and make specific comments.

phantom-duck:
Hello again, and thank you very much for the reply!

We agree that FACTS is not a good fit for `algorithms.postprocessing`. We will move the main method that returns the explanations to `sklearn.explainers` as you suggest.

The related recourse-based fairness metrics can be moved to `sklearn.metrics` as well. Just to be clear, we have 6 of these metrics:
+ Equal Effectiveness
+ Equal Choice for Recourse
+ Equal Effectiveness within Budget
+ Equal Cost of Effectiveness
+ Fair Effectiveness-Cost Trade-Off
+ Equal (Conditional) Mean Recourse

On a higher level, we think FACTS is better stated as a detector. When you choose a detector, the metric of the chosen detector scores the unfairness of all subgroups available and ranks them in decreasing order of their unfairness (more unfair subgroups appear higher in ranking). The reporting that appears is the explainer that explains why the subgroups have the specific unfairness score based on the metric of the detector. 

Looking forward to your view on the matter! In the meantime, we will make sure to rework all of this before the more detailed review you mentioned.

aif360/sklearn/detectors/facts/__init__.py
```diff
@@ -0,0 +1,354 @@
...
+        """
+        if viewpoint == "macro":
+            rules = self.rules_by_if
+        elif viewpoint == "micro":
+            rules = self.rules_with_cumulative
+        else:
+            raise ValueError("viewpoint parameter can be either 'macro' or 'micro'")
+        rules = self.rules_by_if if viewpoint == "macro" else self.rules_with_cumulative
+
+        easy2hard_name_map = {
```
rahulnair23, line 245: Any reason to have a map like this? Why not just use the "easy" names throughout?

aif360/sklearn/detectors/facts/__init__.py
```diff
@@ -0,0 +1,354 @@
...
+        self.rules_by_if = calc_costs(rules_by_if)
+
+        self.dataset = X.copy(deep=True)
+
+        return self
+    
+    def bias_scan(
+        self,
+        metric: str = "equal-effectiveness",
+        viewpoint: str = "macro",
```
rahulnair23, line 156: What does viewpoint mean?

rahulnair23 (commented):
HI @phantom-duck - thanks! This is looking good. Please see some general comments.

hoffmansc:
Thanks for reviewing, @rahulnair23.

One other thing, @phantom-duck: would it be possible to combine the `fit()`, `bias_scan()`, and `print_recourse_report()` methods into a single function without the need for a class? This will better match the detectors API. Right now, I don't really see the point in making three separate calls when only one has outputs. Is it likely that users would run multiple metrics after fitting? The function can have a `verbose` flag to show the progress bars and a `print_recourse_report` flag to show the report otherwise just output the highest scoring subgroup and score like MDSS. If it's easier, you could also just keep the class and implement a wrapper function which does this.

aif360/sklearn/detectors/facts/__init__.py
phantom-duck: In short, due to fear of regression. But you are right, it is not needed, it will be removed.

aif360/sklearn/detectors/facts/__init__.py
phantom-duck: It refers to the notions of "macro viewpoint" and "micro viewpoint" defined in section 2.2 of the [paper](https://arxiv.org/pdf/2306.14978.pdf). As a short explanation, consider a set of actions A and a subgroup (cohort / set of individuals) G. Metrics with the macro viewpoint interpretation are constrained to always apply one action from A to the entire G, while metrics with the micro interpretation are allowed to give each individual in G the min-cost action from A which changes the individual's class.

I should mention, though, that not all combinations of `metric` and `viewpoint` are valid, e.g. "Equal Choice for Recourse" only has a macro interpretation. Do you think this should be reflected in the code somehow?

phantom-duck:
Hello to both! Thank you for the comments, and sorry for the delayed response.

@rahulnair23 Mainly, thank you very much for the detailed review!! For the comments I have replied individually, with the exception of the demo notebook, where the points you raise are interesting and seem trickier to me (especially 2) so I will get back on that soon.

@hoffmansc Ok, thanks for this! Because it is not perfectly clear to me what exactly is the structure of the detectors API. For example, I do not see how the signature of the MDSS function `sklearn.detectors.detectors.bias_scan()` could be applied in our case. The way I see it, the function you describe would need to take as arguments a dataset and the trained model mainly, plus some other more specific parameters. Would something along these lines be okay for what you are thinking?

More importantly, yes, exactly, in our mind, it is entirely possible that the user would run multiple metrics after fitting! Firstly, because fitting takes much longer to run, while the metrics run almost instantaneously. Secondly, because there are several metrics, and with a few extra parameters; so, we would expect (at least some) users to choose a few different overall parameterizations, based on the metrics' descriptions we give in the demo (or in the paper) and domain intuition, and inspect their results. These could very well yield different unfair subgroups.

Due to this, I think it best to keep the class, and of course implement an additional wrapper function as you suggest.

Thank you both again very much!

I will get back to you soon with the modifications and / or further comments. Meanwhile, it goes without saying that we are always available for further discussion.

hoffmansc:
> @hoffmansc Ok, thanks for this! Because it is not perfectly clear to me what exactly is the structure of the detectors API. For example, I do not see how the signature of the MDSS function `sklearn.detectors.detectors.bias_scan()` could be applied in our case. The way I see it, the function you describe would need to take as arguments a dataset `X` and the trained model `clf` mainly, plus some other more specific parameters. Would something along these lines be okay for what you are thinking?

Whatever you need as inputs is fine. I was more referring to the output and function vs. class method(s) paradigm matching MDSS.

As a separate note, is there a more specific/descriptive name than `bias_scan`? Just to differentiate it from the existing detector. Or we could rename that one to `MDSS_bias_scan` and this to `FACTS_bias_scan` or similar, I suppose.

> More importantly, yes, exactly, in our mind, it is entirely possible that the user would run multiple metrics after fitting! Firstly, because fitting takes much longer to run, while the metrics run almost instantaneously. Secondly, because there are several metrics, and with a few extra parameters; so, we would expect (at least some) users to choose a few different overall parameterizations, based on the metrics' descriptions we give in the demo (or in the paper) and domain intuition, and inspect their results. These could very well yield different unfair subgroups.
> 
> Due to this, I think it best to keep the class, and of course implement an additional wrapper function as you suggest.

This makes sense, thanks.

examples/demo_FACTS.ipynb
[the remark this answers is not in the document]
phantom-duck: Hello again!

### Point 1

Ok, I see, you are right, there is a lack of interpretation of the specific experiment's results here. I will add a paragraph with this, and come back to ask whether it is sufficient.

### Point 2

If we are talking on the level of giving some simple suggestions towards the user (such as the ones you mention), then yes, these are the first ideas that would come to mind: targeted mitigations, or perhaps tweak your model's hyperparameters a little and experiment with setups that have suboptimal accuracy but perform better on our fairness metrics. We could mention such suggestions in the demo notebook if you think it would be beneficial.

However, it should be made clear that these suggestions are not part of our work; they would only be general suggestions of the form "try this and run our method again". If we are talking about more than this, then the truth is that it is out of the scope of our work. We only audit fairness, we have not worked on how to mitigate it. Your question is very interesting from this scope! (and perhaps it could be a start for future work on fairness of recourse mitigation) but the fact is that it was not part of our research.

On the more general "how do you expect end users to use this recourse summary", I would say that its intended use was to help the user be *aware* of population subgroups that exhibit *high potential* of recourse-based bias. The rest is up to the user. Which, on one hand, may not be straightforward. But, on the other, we believe that one of the strengths of our framework is its explainability, which means that the user (who possesses domain knowledge) can see the result and immediately understand what is going on and whether it is relevant to them!

PS: If you are talking about [this](https://github.com/AutoFairAthenaRC/AIF360/blob/master/examples/demo_FACTS.ipynb) example, the bias is against females right?

### Point 3

Ok, of course, this is easy to add! The recourse report looks very similar. The difference is only that it also print the message "No Bias". But ok, I am not describing it very well here, so it would be better to add it to the notebook and ask you to check it out.

phantom-duck:
@hoffmansc Hello again!

Ok, that is great, then I will add the wrapper function and get back to you!

As for the name of `bias_scan`, I do not have something much better in mind. The original intention was to much the name of the MDSS function, but in a separate class. Perhaps `find_most_biased_subgroups` or something along those lines would be more informative? Maybe I will think it a bit more. Nonetheless, FACTS_bias_scan seems ok to me. I leave it up to you.

aif360/sklearn/detectors/facts/formatting.py
```diff
@@ -0,0 +1,364 @@
+from typing import List, Dict, Tuple, Optional
+
+import matplotlib.pyplot as plt
+
+from colorama import Fore, Style
```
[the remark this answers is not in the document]
phantom-duck: Done! To be a bit more specific, after running (*note*: only on one machine) the commands:

```shell
conda create -n test --no-shortcuts python<3.9
conda activate test
cd <path/to/AIF360>
pip install -e '.[notebooks, FACTS]'
jupyter lab
```

and opening the demo_FACTS notebook, it is run from start to finish without problems.

aif360/sklearn/detectors/facts/__init__.py
phantom-duck: Done!

phantom-duck:
@hoffmansc The wrapper function is now implemented! I have also added a small snippet of example invocation in the FACTS_demo notebook. Check it out if you please, and inform me of any other changes that need to be made.

PS: for now I have named it `FACTS_bias_scan`, but as I said other suggestions are more than welcome.

examples/demo_FACTS.ipynb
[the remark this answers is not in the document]
phantom-duck: Hello!

Two additional sections have now been added to address points 1 and 3. For point 1, mainly explanatory text has been added. And for point 3 the requested example.

I am a little uncertain as to whether this is sufficient, or you meant that the output of the method itself should be changed to contain more information. Please review it and let me know.

I apologize for the delay on this one. Thank you for your patience!

phantom-duck:
@hoffmansc @rahulnair23 Hello again to both! Just a quick reminder, whenever you have a chance, to review the latest updates in the pull request and let me know of any necessary changes. We are always at your disposal for anything further.

PS: albeit a little late, have a happy and fruitful 2024! :)

rahulnair23:
Looks good to me.

rahulnair23 (approved):

rahulnair23 (approved):

rahulnair23:
Hi @hoffmansc - is this good to merge?

hoffmansc (approved):
other than these tiny things, it looks good to me too.

aif360/sklearn/detectors/facts/__init__.py
```diff
@@ -0,0 +1,418 @@
...
+                - "equal-cost-of-effectiveness"
+                - "equal-mean-recourse"
+                - "fair-tradeoff"
+                
+                Defaults to "equal-effectiveness".
+
+                For explanation of each of those metrics, refer either to the
+                paper [#FACTS23]_ or the demo_FACTS notebook.
+
+            viewpoint (str, optional): "macro" or "micro". Defaults to "macro".
```
hoffmansc, line 249: Add short explanation here.

aif360/sklearn/detectors/__init__.py
```diff
@@ -2,7 +2,9 @@
 Methods for detecting subsets for which a model or dataset is biased.
 """
 from aif360.sklearn.detectors.detectors import bias_scan
+from aif360.sklearn.detectors.facts import FACTS
```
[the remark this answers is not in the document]
phantom-duck: Done.

aif360/sklearn/detectors/facts/__init__.py
phantom-duck: Done. I added the same short description which I also gave in a previous comment. Hopefully, it is sufficient, but I would be happy to make any extra modifications / clarifications you deem appropriate.

phantom-duck:
@hoffmansc 

Hello! That is great, thank you!

I have addressed your small comments. I believe everything should be okay, but as I mentioned above for the 3rd one, perhaps you should take a glimpse to make sure it is sufficient.

Other than that, I have one last remark which I would like to ask you about. The new wrapper method `FACTS_bias_scan` currently lacks any documentation. Until now I assumed this was ok because it simply wraps the `FACTS` class and its methods. Is this true, or should we add a little something?

phantom-duck:
@hoffmansc Hello again!

This is also now done, so I believe everything should be in order. We are looking forward to the addition of our contribution to the aif360 code base! Of course, we are at your disposal for anything further in the meantime.

phantom-duck:
@hoffmansc @rahulnair23 

Hello again. Firstly, thank you both very much for your guidance and cooperation in bringing our contribution to the toolkit!

We also have one question. We would like to be able to provide links to the FACTS section from the official [documentation of aif360](aif360.readthedocs.io). It will be added there eventually right? Would it be possible for you to give an estimation as to when will this happen?

Thank you again very much for everything! :)

hoffmansc:
This should be fixed by #524. See https://aif360.readthedocs.io/en/latest/modules/sklearn.html#module-aif360.sklearn.detectors

Issue #394, opened 2022-09-15
Make notebooks in examples run out of the box using Binder (5/5)  


anupamamurthi:
[tutorial_bias_advertising.ipynb](https://github.com/Trusted-AI/AIF360/blob/master/examples/tutorial_bias_advertising.ipynb)

[tutorial_credit_scoring.ipynb](https://github.com/Trusted-AI/AIF360/blob/master/examples/tutorial_credit_scoring.ipynb)

[tutorial_medical_expenditure.ipynb](https://github.com/Trusted-AI/AIF360/blob/master/examples/tutorial_medical_expenditure.ipynb)

msdokania:
Can I be assigned to this issue? 
I have raised a PR

PR #465, opened 2023-09-22
fix issue #394

siqi-ontheway:
https://github.com/Trusted-AI/AIF360/issues/394


examples/demo_adversarial_debiasing.ipynb
```diff
@@ -1,5 +1,12 @@
 {
  "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
```
gowriaddepalli, line 5: Is there a better way to review `ipynb`?

examples/tutorial_credit_scoring.ipynb
```diff
@@ -55,7 +62,17 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Install AIF360\n",
+    "!pip install 'aif360[lime]'"
```
hoffmansc, line 70: this should just be `!pip install 'aif360'`

hoffmansc: you'll also need to download the german credit score dataset, I believe. See the instructions here:
https://github.com/Trusted-AI/AIF360/tree/master/aif360/data/raw/german

examples/tutorial_medical_expenditure.ipynb
hoffmansc, line 1: same here, you'll need to download the MEPS datasets. This one is a bit more complicated. You should include the disclaimer from [here](https://github.com/Trusted-AI/AIF360/blob/master/aif360/sklearn/datasets/meps_datasets.py) before downloading.

hoffmansc (changes requested):
for all of these, can you run the notebook before committing so it renders the outputs on GitHub

PR #468, opened 2023-09-22
Added a collab badge at the top and also added all the missing aif360

sayalikandarkar:
… dependencies

### DIFF
--- a/examples/demo_exponentiated_gradient_reduction.ipynb
+++ b/examples/demo_exponentiated_gradient_reduction.ipynb
@@ -4,6 +4,8 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


examples/demo_exponentiated_gradient_reduction.ipynb
```diff
@@ -4,6 +4,8 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "[![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://github.com/Trusted-AI/AIF360/blob/master/examples/demo_exponentiated_gradient_reduction.ipynb)\n",
```
gowriaddepalli, line 7: Looks good.

sayalikandarkar: Thank you Gowri! :)

sne3091: The hyperlink here points to the notebook in this repository. The idea of using the badge would be to link it to a colab notebook/asset so it would look something like (`https://colab.research.google.com/<path to notebook>`)

sne3091 (changes requested):

Issue #421, opened 2022-11-28
Broken load_preproc_data_adult


makoeppel:
Running the following code gives me:

```python
from aif360.algorithms.preprocessing.optim_preproc_helpers.data_preproc_functions import load_preproc_data_adult

# Get the dataset and split into train and test
dataset_orig = load_preproc_data_adult()
```

```
Traceback (most recent call last):
  File "/home/makoeppel/ranking/combined_project/notebooks/test.py", line 4, in <module>
    dataset_orig = load_preproc_data_adult()
  File "/home/makoeppel/.local/lib/python3.10/site-packages/aif360/algorithms/preprocessing/optim_preproc_helpers/data_preproc_functions.py", line 77, in load_preproc_data_adult
    return AdultDataset(
  File "/home/makoeppel/.local/lib/python3.10/site-packages/aif360/datasets/adult_dataset.py", line 107, in __init__
    super(AdultDataset, self).__init__(df=df, label_name=label_name,
  File "/home/makoeppel/.local/lib/python3.10/site-packages/aif360/datasets/standard_dataset.py", line 85, in __init__
    df = custom_preprocessing(df)
  File "/home/makoeppel/.local/lib/python3.10/site-packages/aif360/algorithms/preprocessing/optim_preproc_helpers/data_preproc_functions.py", line 13, in custom_preprocessing
    df['Age (decade)'] = df['age'].apply(lambda x: x//10*10)
  File "/home/makoeppel/.local/lib/python3.10/site-packages/pandas/core/series.py", line 4433, in apply
    return SeriesApply(self, func, convert_dtype, args, kwargs).apply()
  File "/home/makoeppel/.local/lib/python3.10/site-packages/pandas/core/apply.py", line 1088, in apply
    return self.apply_standard()
  File "/home/makoeppel/.local/lib/python3.10/site-packages/pandas/core/apply.py", line 1143, in apply_standard
    mapped = lib.map_infer(
  File "pandas/_libs/lib.pyx", line 2870, in pandas._libs.lib.map_infer
  File "/home/makoeppel/.local/lib/python3.10/site-packages/aif360/algorithms/preprocessing/optim_preproc_helpers/data_preproc_functions.py", line 13, in <lambda>
    df['Age (decade)'] = df['age'].apply(lambda x: x//10*10)
TypeError: unsupported operand type(s) for //: 'str' and 'int'
```
My env:
```
aif360==0.5.0
```

Downloaded the datasets from:
[...]

PR #484, opened 2023-09-22
Signed-off-by: Padmaja Borwankar <padmajaborwankar@gmail.com>

padmajaborwankar:
Fixed str to int conversion issue

### DIFF
--- a/aif360/algorithms/preprocessing/optim_preproc_helpers/data_preproc_functions.py
+++ b/aif360/algorithms/preprocessing/optim_preproc_helpers/data_preproc_functions.py
@@ -10,7 +10,7 @@ def load_preproc_data_adult(protected_attributes=None, sub_samp=False, balance=F
         """
 
         # Group age by decade
-        df['Age (decade)'] = df['age'].apply(lambda x: x//10*10)
+        df['Age (decade)'] = df['age'].astype(int).apply(lambda x: x//10*10)
         # df['Age (decade)'] = df['age'].apply(lambda x: np.floor(x/10.0)*10.0)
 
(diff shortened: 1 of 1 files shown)


aif360/algorithms/preprocessing/optim_preproc_helpers/data_preproc_functions.py
```diff
@@ -10,7 +10,7 @@ def custom_preprocessing(df):
         """
 
         # Group age by decade
-        df['Age (decade)'] = df['age'].apply(lambda x: x//10*10)
+        df['Age (decade)'] = df['age'].astype(int).apply(lambda x: x//10*10)
```
gowriaddepalli, line 13: Can we group integers or it is only feasible to group str?

yehee:
It'd be great if you can link to the issue #421 👍

aif360/algorithms/preprocessing/optim_preproc_helpers/data_preproc_functions.py
padmajaborwankar: The 'age' had a String datatype due to which it was typecasted to integer and grouped as a decade.

mnagired (approved):
although I was unable to recreate the bug itself, the changes don't break anything (tested on `aif360==0.5.0` env) so it looks good to me

P.S. when doing `print(type(df['age'].values[0]))` on the line before the comment "Group age by decade", it looks like the type is `<class 'numpy.int64'>`, which explains why the issue wasn't able to be recreated.

Regardless, it's good to explicitly do `astype(int)`

PR #486, opened 2023-09-22
Update tutorial_bias_advertising.ipynb

msdokania:


examples/tutorial_bias_advertising.ipynb
```diff
@@ -1259,4 +1266,4 @@
  },
  "nbformat": 4,
  "nbformat_minor": 5
-}
\ No newline at end of file
+}
```
gowriaddepalli, line 1269: I dont think we need this additional line here

anupamamurthi:
Hi @msdokania

I see that you have added a badge but I am unable to run the notebook.

I see two issues
1. You will need to pip install aif360 as part of the notebook

# Install AIF360
!pip install 'aif360'

2. you will also need "ad_campaign_data.csv" to be downloaded as part of the notebook. I would recommend using a wget or use python's request module to grab bias-in-advertising.tar.gz
3. Finally, u will still need code to unarchive the above file to obtain ad_campaign_data.csv

anupamamurthi (changes requested):
Requesting changes

Issue #378, opened 2022-09-14
Get demo_adversarial_debiasing.ipynb to work in Google Collab


anupamamurthi:

Yashaswini-Viswanath:
Working on this issue

vishnoitanuj:
Hi @hoffmansc Can you please assign me this issue? Want to work on it

divyaamin9825:
Hi can I be assigned this issue?

ranjanadobal:
Hi, could you assign me to this issue?

Srabasti:
Hi, I would like to work on this issue. Since the 1st 2 might be picked up already, picking up the 3rd.

Srabasti:
Hello,
I have checked and confirmed after running, that [demo_grid_search_reduction_classification_sklearn.ipynb](https://github.com/Trusted-AI/AIF360/blob/master/examples/sklearn/demo_grid_search_reduction_classification_sklearn.ipynb) runs without any errors in Colab.
Hence not submitting any PR for this. 
Please let me know for any questions or concerns.
Thanks in advance!
Srabasti Banerjee

PR #493, opened 2023-09-22
Installed all functionalities of aif360

divyaamin9825:
The ExponentiatedGradientFunction code ran differently for different systems if using CPU on colab.
Windows systems could run the code with no issues after installing aif360 with reductions functionality only while M1 Mac systems needed all functionalities of aif360 to run the issues with no errors otherwise it would throw a NameError in the root ExponentiatedGradientFunction code where it couldn't find `import fairlearn.reductions as red`.(Screenshots taken from on an M1 Mac system) 
Screenshot 2023-09-22 at 6 53 02 PM
Screenshot 2023-09-22 at 6 49 08 PM

### DIFF
--- a/examples/sklearn/demo_exponentiated_gradient_reduction_sklearn.ipynb
+++ b/examples/sklearn/demo_exponentiated_gradient_reduction_sklearn.ipynb
@@ -1,997 +1,1753 @@
 {
- "cells": [
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "# Sklearn compatible Exponentiated Gradient Reduction\n",
-    "\n",
-    "Exponentiated gradient reduction is an in-processing technique that reduces fair classification to a sequence of cost-sensitive classification problems, returning a randomized classifier with the lowest empirical error subject to \n",
-    "fair classification constraints. The code for exponentiated gradient reduction wraps the source class \n",
-    "`fairlearn.reductions.ExponentiatedGradient` available in the https://github.com/fairlearn/fairlearn library,\n",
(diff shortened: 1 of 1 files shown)


sne3091:
Contributes towards https://github.com/Trusted-AI/AIF360/issues/378

hoffmansc:
Hi, @divyaamin9825. Thanks for looking into this.

I'd like to get to the bottom of the issue in this case instead of just installing everything. I'm a little confused by the setup. Are both machines, Windows and Mac, running on Colab? If so, shouldn't the operating system not matter? Either way, if there's some difference between OSes, it would be good to know so we can fix the `'Reductions'` requirements.

From the screenshots it looks like you ran the notebook multiple times. Can you try restarting the kernel and installing from scratch?

divyaamin9825:
Hi @hoffmansc. I just ran the code again today with just aif360[Reductions, LawSchoolGPA] and it runs smoothly. However the day of the open source day, installing just those two functionalities was simply not working even in the first run for me on colab. So I had to install all the functionalities. My windows using teammates had their code running smoothly after installing those two packages on colab but I kept encountering the error in the screenshots even after restarting the kernel.

divyaamin9825:
@hoffmansc Hi Samuel, any updates on this?

hoffmansc:
So it works without needing to install `[all]`, right? It should only require `[Reductions]` not `[Reductions,LawSchoolGPA]`. Can you update the PR to reflect this?

Can you also add a "Open in Colab" badge to the top like the other examples (e.g., https://github.com/Trusted-AI/AIF360/blob/master/examples/sklearn/demo_grid_search_reduction_classification_sklearn.ipynb)?

Issue #497, opened 2023-10-03
AttributeError: module 'numpy' has no attribute 'float'.


baraldian:
https://github.com/Trusted-AI/AIF360/blob/2653b7b3ce5b8ae62ac9abfc16b1741ea601b435/aif360/algorithms/preprocessing/optim_preproc_helpers/data_preproc_functions.py#L268C47-L268C65

aif360/algorithms/preprocessing/optim_preproc_helpers/data_preproc_functions.py' , line 268, in <lambda>
    df['age'] = df['age'].apply(lambda x: np.float(x >= 26))

AttributeError: module 'numpy' has no attribute 'float'.
`np.float` was a deprecated alias for the builtin `float`. To avoid this error in existing code, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

hoffmansc:
would you be willing to put together a PR fixing this? I think we can just replace it with `float`, no problem.

baraldian:
> would you be willing to put together a PR fixing this? I think we can just replace it with `float`, no problem.

Done!
Thank you for your suggestion.
Did I do it properly?

PR #498, opened 2023-10-09
Update data_preproc_functions.py to solve conversion to float issue

baraldian:
Solving issue 497 when converting values to float due to deprecation of float function in numpy.

 https://github.com/Trusted-AI/AIF360/issues/497#issue-1923826267 

Using pandas built in functions to optimize time.
Proof:
```
timeit.timeit("df['age'].apply(lambda x: float(x >= 26))", number=10, setup="import pandas as pd;df = pd.DataFrame({'age':range(1000000)})")
2.8678155599627644
timeit.timeit("(df['age'] >= 26).astype(float)", number=10, setup="import pandas as pd;df = pd.DataFrame({'age':range(1000000)})")
0.01317960600135848
((df['age'] >= 26).astype(float) == df['age'].apply(lambda x: float(x >= 26))).all()
True
```


PrimozGodec:
Thank you for the fix. @hoffmansc, can a new release be made soon since it is causing issues on Numpy from 1.24? It is causing conflicts with some packages.

Issue #505, opened 2023-12-14
[Dependencies] module 'numpy' has no attribute 'float'.


HuangruiChu:
`np.float` was a deprecated alias for the builtin `float`. To avoid this error in existing code, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

When I run the code to test with German Dataset, I run into above error.

HuangruiChu:
It seems because the version of AIF360 I installed using "pip install aif360" is different thant current version we have in Github. I just check the code here https://github.com/Trusted-AI/AIF360/blob/master/aif360/algorithms/preprocessing/optim_preproc_helpers/data_preproc_functions.py#L268

This should not cause the above error.

PrimozGodec:
I believe it can be solved with new release, which was already requested long ago but we go no response: https://github.com/Trusted-AI/AIF360/issues/500

PR #521, opened 2024-03-17
Avoiding uses of np.float because that has been removed from the latest numpy.

hirzel:


hirzel:
This is another instance of issue #505.

A good reviewer for this PR would be @hoffmansc.

hoffmansc (approved):

Issue #532, opened 2024-05-20
FACTS method bug: `feature_weights` parameter of the method is not propagated correctly


phantom-duck:
Inside the `fit` method of the `FACTS` detector, the function `calc_costs` is called to calculate the costs of the recourses. But the `params` argument (which includes the `feature_weights`) is not passed. As a result, the default parameters are always used, which assign weight equal to 1 to all features.

Example to reproduce:
```python
from sklearn.model_selection import train_test_split
from sklearn.linear_model import LogisticRegression
from sklearn.compose import ColumnTransformer
from sklearn.pipeline import Pipeline
from sklearn.preprocessing import OneHotEncoder

from aif360.sklearn.datasets.openml_datasets import fetch_adult
from aif360.sklearn.detectors.facts.clean import clean_dataset
from aif360.sklearn.detectors.facts import FACTS
from aif360.sklearn.detectors.facts.predicate import Predicate
import pandas as pd

random_seed = 131313 # to produce the expected if-then clause

# load the adult dataset and perform some simple preprocessing steps
# See output for a glimpse of the final dataset's characteristics
X, y, sample_weight = fetch_adult()
data = clean_dataset(X.assign(income=y), "adult")

# split into train-test data
y = data['income']
X = data.drop('income', axis=1)
X_train, X_test, y_train, y_test = train_test_split(X, y, train_size=0.7, random_state=random_seed, stratify=y)

#### here, we incrementally build the example model. It consists of one preprocessing step,
#### which is to turn categorical features into the respective one-hot encodings, and
#### a simple scikit-learn logistic regressor.
categorical_features = X.select_dtypes(include=["object", "category"]).columns.to_list()
categorical_features_onehot_transformer = ColumnTransformer(
    transformers=[
        ("one-hot-encoder", OneHotEncoder(), categorical_features)
    ],
    remainder="passthrough"
)
model = Pipeline([
    ("one-hot-encoder", categorical_features_onehot_transformer),
    ("clf", LogisticRegression(max_iter=1500))
])

#### train the model
model = model.fit(X_train, y_train)

detector = FACTS(
    clf=model,
    prot_attr="sex",
    freq_itemset_min_supp=0.08,
    feature_weights={f: 10 for f in X.columns},
    feats_not_allowed_to_change=[]
)

detector = detector.fit(X_test)

[...]

PR #533, opened 2024-05-20
Bug fixes for the FACTS method

### LINKED ISSUE #531: FACTS method bug: code in `valid_ifthens` function has hard-coded feature names
In the internal `valid_ifthens` function, there are 2 points where there exist hard-coded feature names in the code, and these parts of the code fail without them. Specifically:

1. [here](https://github.com/Trusted-AI/AIF360/blob/7c4f172f4a81b1d14d19400e95c468aaf3134952/aif360/sklearn/detectors/facts/misc.py#L285), there is a reference to an `age` column of `X`, which is also assumed to be of `pd.Interval` dtype. As such, if this column does not exist or is not of Interval dtype, this part of the code throws an error.

Example to reproduce:
```python
from sklearn.model_selection import train_test_split
from sklearn.linear_model import LogisticRegression
from sklearn.compose import ColumnTransformer
from sklearn.pipeline import Pipeline
from sklearn.preprocessing import OneHotEncoder

from aif360.sklearn.datasets.openml_datasets import fetch_german
from aif360.sklearn.detectors.facts import FACTS

X, y = fetch_german()
assert (X.index == y.index).all()
X.reset_index(drop=True, inplace=True)
y = y.reset_index(drop=True).map({"bad": 0, "good": 1})

# split into train-test data
X_train, X_test, y_train, y_test = train_test_split(X, y, train_size=0.7, stratify=y)

categorical_features = X.select_dtypes(include=["object", "category"]).columns.to_list()
categorical_features_onehot_transformer = ColumnTransformer(
    transformers=[
        ("one-hot-encoder", OneHotEncoder(), categorical_features)
    ],
    remainder="passthrough"
)
model = Pipeline([
    ("one-hot-encoder", categorical_features_onehot_transformer),
    ("clf", LogisticRegression(max_iter=1500))
])

#### train the model
model = model.fit(X_train, y_train)

detector = FACTS(
    clf=model,
    prot_attr="sex",
    feature_weights={f: 1 for f in X.columns},
    feats_not_allowed_to_change=[]
)

detector = detector.fit(X_test)
```

The last command fails with `AttributeError: 'numpy.float64' object has no attribute 'left'`

2. At [this point](https://github.com/Trusted-AI/AIF360/blob/7c4f172f4a81b1d14d19400e95c468aaf3134952/aif360/sklearn/detectors/facts/misc.py#L269), the `recIsValid` function is used. This, in turn, [here](https://github.com/Trusted-AI/AIF360/blob/7c4f172f4a81b1d14d19400e95c468aaf3134952/aif360/sklearn/detectors/facts/predicate.py#L198) also references hard-coded feature names. Here there exist checks of whether they exist or not, so the code does not fail if they do not exist. But there are cases where if a feature exists, it is assumed either to be of a certain type or to possess certain semantics.

I do not currently have a reproducible example for this one, because whether it will appear or not depends on the exact test data. I believe, however, that it is clear this is also a bug, and if we want to enforce some constraints, such as this part of the code is trying to do, it should be done in some other, more robust way.

phantom-duck:
@hoffmansc 

Wrote fixes for the bugs documented in issues #531 and #532.

For the issue of the hard-coded features in the `recIsValid` function (issue #531, code [here](https://github.com/Trusted-AI/AIF360/blob/7c4f172f4a81b1d14d19400e95c468aaf3134952/aif360/sklearn/detectors/facts/predicate.py#L198)), the fix currently proposed is a hotfix. The values have not been removed, only the relevant argument's default value has been changed from `True` to `False`.

Whether you would like us to remove this part of the code altogether or keep working on it to make it more robust (and if so, in this pull request or a new one), please let me know.

Fixes #531 and #532.

### DIFF
--- a/aif360/sklearn/detectors/facts/__init__.py
+++ b/aif360/sklearn/detectors/facts/__init__.py
@@ -359,7 +359,7 @@ class FACTS(BaseEstimator):
             params=params,
             verbose=verbose,
         )
-        self.rules_by_if = calc_costs(rules_by_if)
+        self.rules_by_if = calc_costs(rules_by_if, params=params)
 
         self.dataset = X.copy(deep=True)
 
--- a/aif360/sklearn/detectors/facts/__init__.py
+++ b/aif360/sklearn/detectors/facts/__init__.py
@@ -342,6 +342,7 @@ class FACTS(BaseEstimator):
             model=self.clf,
             sensitive_attribute=self.prot_attr,
             freqitem_minsupp=self.freq_itemset_min_supp,
+            drop_infeasible=False,
             feats_not_allowed_to_change=list(feats_not_allowed_to_change),
             verbose=verbose,
         )
--- a/aif360/sklearn/detectors/facts/misc.py
+++ b/aif360/sklearn/detectors/facts/misc.py
@@ -6,7 +6,7 @@ import numpy as np
 from pandas import DataFrame
 
 from .parameters import *
-from .predicate import Predicate, recIsValid, featureChangePred, drop_two_above
+from .predicate import Predicate, recIsValid, featureChangePred
 from .frequent_itemsets import run_fpgrowth, preprocessDataset, fpgrowth_out_to_predicate_list
 from .metrics import (
     incorrectRecoursesIfThen,
@@ -182,7 +182,6 @@ def valid_ifthens(
     freqitem_minsupp: float = 0.01,
     missing_subgroup_val: str = "N/A",
     drop_infeasible: bool = True,
-    drop_above: bool = True,
     feats_not_allowed_to_change: List[str] = [],
     verbose: bool = True,
 ) -> List[Tuple[Predicate, Predicate, Dict[str, float], Dict[str, float]]]:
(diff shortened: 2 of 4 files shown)


hoffmansc (approved):
This PR is good for now. For #531, it would be good to have a more robust system eventually. For #532, can we add a short test case like in the issue to show it's working?

Thanks!

phantom-duck:
Hello! Thanks for the feedback. Apologies for the delayed response.

> For https://github.com/Trusted-AI/AIF360/issues/531, it would be good to have a more robust system eventually

Ok, great, we will see to it as soon as possible and create a new PR.

> For https://github.com/Trusted-AI/AIF360/issues/532, can we add a short test case like in the issue to show it's working?

Of course. So this one I will add to this PR right?

hoffmansc:
> Of course. So this one I will add to this PR right?

Yes, please 🙂

phantom-duck:
Hello again @hoffmansc !

I have now added a test showcasing that the costs are computed correctly (close to the example in #532). Just a question to be sure I understood correctly, is this what you meant? Or by "test case" you mean also to add something to the example notebook for example?

Other than, I think we are good to go.

hoffmansc:
This is perfect, thanks!

Issue #529, opened 2024-04-24
Demo notebook needs update 


Shidhanta95:
Example jupyter notebook for sklearn (https://github.com/Trusted-AI/AIF360/blob/main/examples/sklearn/demo_new_features.ipynb) throws an error during compilation because of an outdated usage of the OneHotEncoder usage.  The fix needed is for parameter to be updated from sparse to sparse_output.

Shidhanta95:
If allowed to, I can make the necessary changes and make a pull request as well. Just wanted to document the issue.

codell2:
I have worked on this issue and attached the PR above with the fix.

PR #541, opened 2024-12-03
Fix OneHotEncoder usage in sklearn example notebook

codell2:
Updated the `OneHotEncoder` usage in `examples/sklearn/demo_new_features.ipynb` to replace the deprecated `sparse` parameter with `sparse_output`. This resolves an error on Issue #529 during compilation due to recent updates in sklearn's API.

### DIFF
--- a/examples/sklearn/demo_new_features.ipynb
+++ b/examples/sklearn/demo_new_features.ipynb
@@ -632,7 +632,7 @@
    ],
    "source": [
     "ohe = make_column_transformer(\n",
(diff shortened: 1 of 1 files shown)


codell2:
The test suite is failing due to an issue in `tests/sklearn/test_datasets.py` involving the use of `np.NaN`, which was removed in NumPy 2.0. This issue is unrelated to my change, which only updates `demo_new_features.ipynb`. The issue can be fixed by changing this to `np.nan`.

